PR #2, opened 2021-08-24
Introduce inactive state for partition devices

weihanglo:
# Introduce inactive state for partition devices

## Background

When we adding a new partition, more than one uevents of that partition are generated, which in the sequence: add, remove, add. This behaviour makes it difficult to blindly delete the blockdevice CR when receiving an uevent with remove action. 

## Proposal

We propose not to remove any parition. We only flag partitions inactive when receiving uevent with remove action against them. The actual removal of a partition device only happens when we tried to remove the whole disk.

## Test plan

At this moment we need to manually test it.

1. Create a k8s cluster with longhorn installed.
2. Run the binary. e.g.`go run main.go`. Don't forget to specify `--namespace` and `--node-name`.
3. Watch the changes of all blockdevice CR. 
3. Plug a raw disk (without any partition) on the machine
    - Expect to see two blockdevices if you cannot generate a valid GUID from the disk: one with deviceType Disk; the other with deviceType Part
    - Expect to see one blockdevice with disk type if your disk has WWN or S/N or GPT partition table inside.
4. Edit the partition blockdevice's `Spec.FileSystem.MountPoint` to a valid path. and also set `Spec.FileSystem.ForceFormatted` to true. ⚠️ This would format the disk
5. You are expected to see that the state of the partition device transit from active to inactive and then back to active.

### DIFF
--- a/pkg/udev/uevent.go
+++ b/pkg/udev/uevent.go
@@ -125,6 +125,8 @@ func (u *Udev) ActionHandler(uevent netlink.UEvent) {
 	case netlink.REMOVE:
 		if udevDevice.IsDisk() {
 			u.RemoveBlockDevice(bd, &udevDevice, disk, defaultDuration)
+		} else if udevDevice.IsPartition() {
+			u.deactivateBlockDevice(bd)
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/udev/uevent.go
```diff
@@ -173,11 +175,32 @@ func (u *Udev) RemoveBlockDevice(device *v1beta1.BlockDevice, udevDevice *Device
...
+
+		logrus.Errorf("failed to deactivate block device %s, error: %s", device.Name, err.Error())
+		return
+	}
+
+	logrus.Debugf("deactivate block deivce %s on path %s", bd.Name, bd.Spec.DevPath)
+
+	deviceCpy := bd.DeepCopy()
+	deviceCpy.Status.State = v1beta1.BlockDeviceInactive
+	if _, err := u.controller.Blockdevices.Update(deviceCpy); err != nil && !errors.IsNotFound(err) {
```
gitlawr, line 199: Can we skip the update if the block device state is already inactive?

weihanglo: Of course! Fixed on commit 468c8b6. Thanks for your suggestion 😄

bk201 (approved):

PR #3, opened 2021-08-26
Tests for utility functions and adding/removing disk for longhorn nodes

weihanglo:
Tests includes:

- Device GUID generating Logic (for both disk and partition)
- Disk filter logic (for both vendor filter and path filter)
- Controller logic of adding/removing additional disk for Longhorn node when specifying `blockdevice.spec.mountPoint`
- Logic of `parseMountEntry`

This PR also did the following stuff as well:

- Introduce new dependency [github.com/stretchr/testify](https://pkg.go.dev/github.com/stretchr/testify)
- Add `fakeclients` as test doubles for blockdevice and node

Node that these PR still does not yet cover all critical logic in node-disk manager. We'll add more tests and integrate CI pipeline in other follow ups.


guangbochen:
@weihanglo can u rebase this PR and review this as the unit test is a always nice thing to have, thanks.

weihanglo:
Some of tests here will be re-introduced in #17. I plan to rebase this after #17. Thanks.

PR #4, opened 2021-08-30
Switch image repo to rancher/harvester-node-disk-manager

weihanglo:
As title, we publish the docker image under repo `rancher/harvester-node-disk-manager` at this moment and remove arm support temporarily.


guangbochen (approved):
LGTM

PR #5, opened 2021-09-06
Periodic rescanning for blockdevices

weihanglo:
## What

Add periodic rescanning for blockdevices
                                                                     
- A new CLI flag `--rescan-interval`. Default to 60 seconds (feel free to discuss the default in comments)
- A simple periodic blockdevice rescanning mechanism that reuses the
  logic from good old `RegisterNodeBlockDevices` function

## Test plan 

1. Create a k8s cluster with longhorn installed.
2. Run the binary. e.g.`go run main.go`. Don't forget to specify `--namespace` and `--node-name`.
    - You can add `--rescan-interval` with a small value, say `10` seconds, to watch the changes more efficiently.
3. Watch the changes of all blockdevice CR. 
3. Plug a raw disk (without any partition) on the machine
    - Expect to see some block device CR being created.
4. Manually create or delete GPT partitions
    - Expect to see the block device CR being created and delete at each rescan interval.

### DIFF
--- a/main.go
+++ b/main.go
@@ -105,6 +105,7 @@ func main() {
 		},
 		&cli.Int64Flag{
 			Name:        "rescan-interval",
+			EnvVars:     []string{"NDM_RESCAN_INTERVAL"},
 			Usage:       "Specify the interval of device rescanning of the node (in seconds)",
 			Destination: &opt.RescanInterval,
 		},
--- a/main.go
+++ b/main.go
@@ -109,6 +109,12 @@ func main() {
 			Usage:       "Specify the interval of device rescanning of the node (in seconds)",
 			Destination: &opt.RescanInterval,
 		},
+		&cli.BoolFlag{
+			Name:        "auto-gpt-generate",
+			EnvVars:     []string{"NDM_AUTO_GPT_GENERATE"},
+			Usage:       "Enable auto GPT partition generating if a disk can not be globally identified",
+			Destination: &opt.AutoGPTGenerate,
+		},
 	}
 
 	app.Action = func(c *cli.Context) error {
--- a/main.go
+++ b/main.go
@@ -103,6 +103,11 @@ func main() {
 			Usage:       "A string of comma-separated values that you want to exclude for block device path filter",
 			Destination: &opt.PathFilter,
 		},
+		&cli.Int64Flag{
+			Name:        "rescan-interval",
+			Usage:       "Specify the interval of device rescanning of the node (in seconds)",
+			Destination: &opt.RescanInterval,
+		},
 	}
 
 	app.Action = func(c *cli.Context) error {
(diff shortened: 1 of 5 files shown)


guangbochen (approved):
LGTM

main.go
```diff
@@ -103,6 +103,11 @@ func main() {
 			Usage:       "A string of comma-separated values that you want to exclude for block device path filter",
 			Destination: &opt.PathFilter,
 		},
+		&cli.Int64Flag{
+			Name:        "rescan-interval",
+			Usage:       "Specify the interval of device rescanning of the node (in seconds)",
+			Destination: &opt.RescanInterval,
+		},
```
bk201, line 110: nit: will be nice to support EnvVars too. Easier to use than args in some cases.

weihanglo: @bk201
fixed at a518a88, thanks.

pkg/controller/blockdevice/controller.go
```diff
@@ -76,11 +97,21 @@ func (c *Controller) RegisterNodeBlockDevices() error {
 			continue
 		}
 
-		logrus.Infof("Found a block device /dev/%s", disk.Name)
+		logrus.Debugf("Found a disk block device /dev/%s", disk.Name)
+
 		bd := GetDiskBlockDevice(disk, c.nodeName, c.namespace)
+
+		if newBd, err := c.MakeGPTPartitionIfNeeded(bd); err != nil {
```
bk201, line 104: This will automatically create GPT partition on un-identified disks (e.g. virtio disks).
We better have an option to enable/disable this feature.

bk201 (changes requested):

pkg/controller/blockdevice/controller.go
weihanglo: Good suggestion. It's also straightforward to add such an option. The questions are:

1. If this feature is disabled, what is the expected behaviour? IMO throwing out the error directly is a good choice.
2. Should this option be enabled by default? If this defaults to enabled, then I guess people whould keep it as is until they hit the wall of auto-create-gpt-partition.

bk201: Logging an error is good. Will a BlockDevice resource be created in this situation? If yes, maybe setting conditions also works.

I'd recommend disabling the feature by default. Do not touch users' disks unless they agree.

weihanglo: No. If it cannot identify a disk, a block device won't be generated.
I think logging errors and disabling the feature by default are great.

cc @guangbochen. What do you think?

guangbochen: Yep, LGTM, IIRC when the user mounts the entire single block device to the node, we will create the GPT table and partition of that disk, correct?

weihanglo: Correct. I just pushed a commit f629cca for this new option. Thanks for the suggestion @bk201

bk201 (approved):
LGTM! Thanks!

PR #6, opened 2021-09-07
Call mkdir only when diretory does not exist

weihanglo:
## Problem

To mount a new device on an existing directory of a read-only filesystem, NDM should not require the file system to be read-write.

## Solution

We inspect the existence of the directory first to decide whether NDM need to execute `mkdir -p` or not.


guangbochen (approved):
LGTM

bk201 (approved):

PR #7, opened 2021-09-08
Remove leader since NDM should run on every nodes

weihanglo:
NDM as a daemonset should run on every nodes to detect disk attachments
and detachments. Therefore the leader election is unnecessary for NDM.


guangbochen (approved):
LGTM

PR #8, opened 2021-09-08
Perform operations only on the node NDM located

weihanglo:
## Problem

When attaching/detaching/provisioning device from one node, it deletes blockdevices on other nodes.

## Approaches

In order to let each NDM instance perform operations only on its own node, we did following changes:

- Filter blockdevice by label "kubernetes.io/hostname" on blockdevice to
  avoid NDM from accidentally modifying other nodes' resources.
- Provide valid args to construct blockdevice.controller for udev
  We need to provide all valid args to blockdevice.controller in order to
  let it work correctly. Therefore we need to
    - Expose all fields of `blockdevice.controller`
    - Modify `udev.NewUdev` signature to accept longhorn Node controller
- NDM now only concern blockdevice on the same node.
- Remove leader election (copied from #7)

This PR also fixed the update loop as well (11e05ac)

### DIFF
--- a/pkg/controller/blockdevice/controller.go
+++ b/pkg/controller/blockdevice/controller.go
@@ -159,8 +159,8 @@ func (c *Controller) ScanBlockDevicesOnNode() error {
 // OnBlockDeviceChange watch the block device CR on change and performing disk operations
 // like mounting the disks to a desired path via ext4
 func (c *Controller) OnBlockDeviceChange(key string, device *diskv1.BlockDevice) (*diskv1.BlockDevice, error) {
-	if device == nil || device.DeletionTimestamp != nil {
-		return device, nil
+	if device == nil || device.DeletionTimestamp != nil || device.Spec.NodeName != c.NodeName {
+		return nil, nil
 	}
 
 	deviceCpy := device.DeepCopy()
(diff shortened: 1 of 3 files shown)


pkg/controller/blockdevice/controller.go
```diff
@@ -126,7 +127,11 @@ func (c *Controller) ScanBlockDevicesOnNode() error {
 		}
 	}
 
-	oldBdList, err := c.Blockdevices.List(c.namespace, v1.ListOptions{})
+	oldBdList, err := c.Blockdevices.List(c.namespace, v1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{
+			corev1.LabelHostname: c.nodeName,
+		}).String(),
+	})
```
bk201, line 134: Maybe just:
```suggestion
	oldBdList, err := c.Blockdevices.List(c.namespace, v1.ListOptions{
		LabelSelector: fmt.Sprintf("%s=%s", corev1.LabelHostname, c.nodeName),
	})
```

guangbochen (approved):

gitlawr (approved):
LGTM

bk201 (approved):
LGTM. Disks are created correctly.

PR #9, opened 2021-09-08
Let SaveBlockDevice update more precisely

weihanglo:
A new block device status is derived from sysfs, and that results
in lacking `FileSystem.LastFormattedAt`. We copy that field before
`DeepEqual` comparison to reduce unnecessary updates.


guangbochen (approved):
LGTM

PR #10, opened 2021-09-10
Node scoped device GUID and use blake2b instead of md5

weihanglo:
Address https://github.com/harvester/harvester/issues/1195 and https://github.com/harvester/harvester/issues/1196

- Add `nodeName` to hash function, so the same devices on different nodes are treated as different block device.
- Use blake2b instead of md5 to produce hash value. I am not sure whether 16-bytes digest size is ennough, but there is a limit that a label must not be more than 63 characters, so we cannot use the recommended size (32-bytes) of blake2b hash.


harvester-node-disk-manager/Chart.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: v2
```
guangbochen, line 1: ditto

harvester-node-disk-manager/.helmignore
```diff
@@ -0,0 +1,23 @@
+# Patterns to ignore when building packages.
```
guangbochen, line 1: why do we need this file?

harvester-node-disk-manager/Chart.yaml
weihanglo: A mistake. Sorry I'll remove them.

harvester-node-disk-manager/.helmignore
weihanglo: Removed. Sorry for the inconvenience.

guangbochen (approved):
LGTM

bk201 (approved):
LGTM!

PR #12, opened 2021-09-27
Able to exclude partition based on ext filesystem label

weihanglo:
Related issue: https://github.com/harvester/harvester/issues/1282

- Separate filter to two filter interface: DiskFilter and PartFilter
- Add new label filter which excludes partition based on given pattern of filesystem label
- Add new cli arg `--label-filter` and new environment variable `NDM_LABEL_FILTER`. It accept a comma separated glob patterns describing in [`path/filepath#Match`](https://pkg.go.dev/path/filepath#Match)

## Following tasks after merged.

- [ ] Update helm chart for harvester/charts repo https://github.com/harvester/charts/pull/34
- [ ] For harvester-installer, add a new value to exclude `COS_*` label https://github.com/harvester/harvester-installer/pull/140

## Test plan

> You can test the docker image at [`weihanglo/node-disk-manager:label-filter`](https://hub.docker.com/layers/169481236/weihanglo/node-disk-manager/label-filter/images/sha256-8a73fe7d2bf56f6b4fa383154f5eb90f097a8838c956080df89825f833d96f45)

1. Run node-disk-manager with `--label-filter` arg, e.g. `--label-filter "COS_*"` to exclude partition with ext label starting with `COS_` prefix
2. Label your partition with that label, e.g. `e2label /dev/sda1 MY_LABEL_*`
3. You shall see that periodic scanning removes partitions matching that filter value.

### DIFF
--- a/main.go
+++ b/main.go
@@ -101,6 +101,12 @@ func main() {
 			Usage:       "A string of comma-separated values that you want to exclude for block device path filter",
 			Destination: &opt.PathFilter,
 		},
+		&cli.StringFlag{
+			Name:        "label-filter",
+			EnvVars:     []string{"NDM_LABEL_FILTER"},
+			Usage:       "A string of comma-separated glob pattern that you want to exclude for block device filesystem label filter",
+			Destination: &opt.LabelFilter,
+		},
 		&cli.Int64Flag{
 			Name:        "rescan-interval",
 			EnvVars:     []string{"NDM_RESCAN_INTERVAL"},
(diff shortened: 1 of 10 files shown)


bk201:
LGTM. It might be better to filter out the disk that contains the unwanted partitions too.

guangbochen (commented):
for now, I think we can filter out the parent disk if all of its partitions contain the filter label.

weihanglo:
@guangbochen 

Just added a logic for applying label filter on disk. If all of its partitions are excluded, the disk would be excluded, too.

Another fix is that in previous implementation, we exclude the whole disk if one of its partition mounted on the `excludePaths`. That is too restricted, so I separate the logic for each filter to take care only given disk or partition. If you think this change is not what we want, we can revert it. Thanks.

PR #13, opened 2021-09-27
Do not provision raw disk after force formatting

weihanglo:
Related to https://github.com/harvester/harvester/issues/1285

After force formatting a raw disk (e.g. without any partition such as `/dev/sda`), set `spec.fileSystem.provisioned` to false as we do want to provision it single root partition instead of the raw disk itself.

### DIFF
--- a/pkg/controller/blockdevice/controller.go
+++ b/pkg/controller/blockdevice/controller.go
@@ -126,7 +126,7 @@ func (c *Controller) ScanBlockDevicesOnNode() error {
 		for _, part := range disk.Partitions {
 			logrus.Debugf("Found a partition block device /dev/%s", part.Name)
 			bd := GetPartitionBlockDevice(part, c.NodeName, c.Namespace)
-			if len(bd.Name) == 0 {
+			if bd.Name == "" {
 				logrus.Infof("Skip adding non-identifiable block device %s", bd.Spec.DevPath)
 				continue
 			}
@@ -318,7 +318,7 @@ func updateDeviceMount(devPath, mountPoint, existingMount string) error {
 // Currently only making GPT partition on devices without a name (GUID).
 func (c *Controller) MakeGPTPartitionIfNeeded(device *diskv1.BlockDevice) (*diskv1.BlockDevice, error) {
 	devPath := device.Spec.DevPath
-	guidMissing := len(device.ObjectMeta.Name) == 0
+	guidMissing := device.ObjectMeta.Name == ""
 
 	if !guidMissing {
 		// No need to generate new GPT partition if this device is identifiable.
--- a/pkg/controller/blockdevice/controller.go
+++ b/pkg/controller/blockdevice/controller.go
@@ -123,6 +123,10 @@ func (c *Controller) ScanBlockDevicesOnNode() error {
 		for _, part := range disk.Partitions {
 			logrus.Debugf("Found a partition block device /dev/%s", part.Name)
 			bd := GetPartitionBlockDevice(part, c.NodeName, c.Namespace)
+			if len(bd.Name) == 0 {
+				logrus.Infof("Skip adding non-identifiable block device %s", bd.Spec.DevPath)
+				continue
+			}
 			newBds = append(newBds, bd)
 		}
 	}
--- a/pkg/controller/blockdevice/controller.go
+++ b/pkg/controller/blockdevice/controller.go
@@ -15,6 +15,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	diskv1 "github.com/harvester/node-disk-manager/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/node-disk-manager/pkg/block"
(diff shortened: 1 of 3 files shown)


weihanglo:
## Test plan

1. Run node disk manager
1. Create a disk with partions on it
2. Attach the disk
3. Set spec on the disk 
    - `spec.fileSystem.provisioned: true`
    - `spec.fileSystem.forceFormatted: true
    - Give it a mountpoint via `spec.fileSystem.mountPoint`
4. Expect node disk manager create a new partition blockdevice CR

guangbochen (approved):
<strike>LGTM</strike>

weihanglo:
Oh there is a duplicated umount in c51ba3b that I need to remove it. It won't harm existing logic.

bk201:
I'm getting this by force-formatting a disk that already contains a partition (just `sgdisk -n 1:0:0 on the disk, not creating any file system on it):
Screen Shot 2021-09-27 at 11 16 19 PM

node2 log: 
[node2-ndm.log](https://github.com/harvester/node-disk-manager/files/7237620/node2-ndm.log)

condition of longhorn nodes:
```
    diskStatus:
      2730352661b4b0077c11747df84361cb:
        conditions:
          Ready:
            lastProbeTime: ""
            lastTransitionTime: "2021-09-27T15:14:19Z"
            message: 'Disk 2730352661b4b0077c11747df84361cb(/usr/local/lh/sda) on
              node node2 is not ready: cannot find disk config file, maybe due to
              a mount error'
            reason: DiskFilesystemChanged
            status: "False"
            type: Ready
          Schedulable:
            lastProbeTime: ""
            lastTransitionTime: "2021-09-27T15:14:19Z"
            message: the disk 2730352661b4b0077c11747df84361cb(/usr/local/lh/sda)
              on the node node2 is not ready
            reason: DiskNotReady
            status: "False"
            type: Schedulable
```

```
node2:~ # ls /usr/local/lh/sda/
lost+found
node2:~ # df /dev/sda1
Filesystem     1K-blocks  Used Available Use% Mounted on
/dev/sda1       10216724    24   9676132   1% /usr/local/lh/sda
```

Support bundle:
[supportbundle_7346128e-12d2-4a05-a955-0cf529d1f227_2021-09-27T15-46-45Z.zip](https://github.com/harvester/node-disk-manager/files/7237645/supportbundle_7346128e-12d2-4a05-a955-0cf529d1f227_2021-09-27T15-46-45Z.zip)

weihanglo:
This is what c8d5abc did so far.
 
For current GUID generating logic, a raw block device without WWN but
with a GPT partition table will generate a GUID based on it's `PtUUID`.
That means everytime we `forceFormatted` a raw block device, we are
creating another new unique blockdevice from NDM's perspective.

Although NDM has a recurring scan logic to reconcile states of all
blockdevices, I think the best practice is cleaning-up the old device
ASAP, so we manually remove and create a new one right after creating
the new partition block device.

weihanglo:
@guangbochen 

Updated.
- Default rescan interval to 30s
- Umount disk when user tries to unprovision it

guangbochen (approved):

pkg/controller/blockdevice/controller.go
```diff
@@ -123,6 +126,10 @@ func (c *Controller) ScanBlockDevicesOnNode() error {
 		for _, part := range disk.Partitions {
 			logrus.Debugf("Found a partition block device /dev/%s", part.Name)
 			bd := GetPartitionBlockDevice(part, c.NodeName, c.Namespace)
+			if len(bd.Name) == 0 {
```
gitlawr, line 129: `bd.Name == ""` is more straightforward

pkg/controller/blockdevice/controller.go
```diff
@@ -348,59 +367,97 @@ func (c *Controller) forceFormatDisk(device *diskv1.BlockDevice) (*diskv1.BlockD
...
+	// Update the single partition block device to start force formatting
+	toUpdate := bd.DeepCopy()
+	toUpdate.Spec.FileSystem.Provisioned = filesystem.Provisioned
+	toUpdate.Spec.FileSystem.MountPoint = filesystem.MountPoint
+	toUpdate.Spec.FileSystem.ForceFormatted = true
+	diskv1.DeviceFormatting.SetStatusBool(toUpdate, true)
+	diskv1.DeviceFormatting.Message(toUpdate, fmt.Sprintf("formatting disk partition %s with ext4 filesystem", toUpdate.Spec.DevPath))
+	if _, err := c.Blockdevices.Update(toUpdate); err != nil {
+		logrus.Errorf("failed to update partition block device %s, %s", bd.Name, err.Error())
+		return device, err
```
gitlawr, line 434: Don't log the error if we are gonna return it. Expect the caller to handle/log it.

pkg/controller/blockdevice/controller.go
```diff
@@ -348,59 +364,97 @@ func (c *Controller) forceFormatDisk(device *diskv1.BlockDevice) (*diskv1.BlockD
...
+		return device, fmt.Errorf("device %s is not a disk", device.Name)
+	}
+
+	filesystem := device.Spec.FileSystem
+	fsStatus := device.Status.DeviceStatus.FileSystem
+	logrus.Infof("performing format operation of disk %s, mount path %s", device.Spec.DevPath, filesystem.MountPoint)
+
+	// umount the disk if it is mounted
+	if fsStatus.MountPoint != "" {
+		if err := disk.UmountDisk(fsStatus.MountPoint); err != nil {
```
gitlawr, line 397: How/Where do we update the MountPoint status after we umount the disk?

weihanglo: Nearly the end of the controller loop. In fact it is the same logic as before.

https://github.com/harvester/node-disk-manager/blob/d979d56842ef2a58af7cf1166cd339e885701206/pkg/controller/blockdevice/controller.go#L237-L239

pkg/controller/blockdevice/controller.go
weihanglo: fixed with ddf7817

bk201:
I didn't see this [issue](https://github.com/harvester/node-disk-manager/pull/13#issuecomment-928002682) anymore.

bk201 (approved):

PR #15, opened 2021-10-01
Filesystem status must reflect values from sysfs

weihanglo:
Status should be mountpoint from sysfs not the original spec.

Signed-off-by: Weihang Lo <weihang.lo@suse.com>


bk201 (approved):
LGTM

PR #16, opened 2021-10-11
Fix for invalid partition device paths with NVMe disks

janeczku:
See https://github.com/harvester/harvester/issues/1414

### DIFF
--- a/pkg/controller/blockdevice/controller.go
+++ b/pkg/controller/blockdevice/controller.go
@@ -24,6 +24,7 @@ import (
 	ctldiskv1 "github.com/harvester/node-disk-manager/pkg/generated/controllers/harvesterhci.io/v1beta1"
 	ctllonghornv1 "github.com/harvester/node-disk-manager/pkg/generated/controllers/longhorn.io/v1beta1"
 	"github.com/harvester/node-disk-manager/pkg/option"
+	"github.com/harvester/node-disk-manager/pkg/util"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/util/util.go
```diff
@@ -29,6 +30,18 @@ func GetFullDevPath(shortPath string) string {
 	return fmt.Sprintf("/dev/%s", shortPath)
 }
 
+func GetDiskPartitionPath(devicePath string, partitionNum int) string {
+	partitionSep := ""
+	last := devicePath[len(devicePath)-1:]
+	if _, err := strconv.Atoi(last); err == nil {
```
weihanglo, line 36:
```suggestion
func GetDiskPartitionPath(devicePath, storageController string, partitionNum int) string {
	partitionSep := ""
	switch storageController {
	case "NVMe":
```

pkg/controller/blockdevice/controller.go
```diff
@@ -411,7 +412,7 @@ func (c *Controller) forceFormatDisk(device *diskv1.BlockDevice) (*diskv1.BlockD
 	var bd *diskv1.BlockDevice
 	poll := func() (bool, error) {
 		var err error
-		devPath := device.Spec.DevPath + "1"
+		devPath := util.GetDiskPartitionPath(device.Spec.DevPath, 1)
```
weihanglo, line 415:
```suggestion
		storageController := device.Status.DeviceStatus.Details.StorageController
		devPath := util.GetDiskPartitionPath(device.Spec.DevPath, storageController, 1)
```

pkg/util/util.go
janeczku: The current logic mirrors how Linux Kernel generates disk partition names. It's based on trailing digit, not specific disk controller. See:
https://github.com/analogdevicesinc/linux/blob/master/block/partition-generic.c#L39-L40

So maybe it's better to leave that way?

bk201 (approved):

PR #17, opened 2021-10-29
Auto provision for disks

weihanglo:
## What

Add auto-provision command line argument for harvester/harvester#1241

- [x] Add unit tests for filters
- [x] Add device path filter
- [x] Add `--auto-provision-filters` argument

### Provision Hotplug Disks

1. Has a cluster with Longhorn installed
2. Run `node-disk-manager` with disk you want to provision. Take `/dev/vda` for example, run `node-disk-manager --auto-provision-filters "/dev/vda"`
3. Hotplug the disk and you'll see `/dev/vda1` being provisioned as a disk on longhorn node. Use `kubectl -n longhorn-system get lhn -oyaml` to check if the disk exists.

### Provision Existing Disks

1. Has a cluster with Longhorn installed
2. Plug-in a disk you want to provision in advance. For example `/dev/vda`.
2. Run `node-disk-manager --auto-provision-filters "/dev/vda"`
3. You'll see `/dev/vda1` being provisioned as a disk on longhorn node. Use `kubectl -n longhorn-system get lhn -oyaml` to check if the disk exists.


johnliu55tw:
@weihanglo I think we also need the `--force-format-disk` option? For forcefully formatting the disk regardless it already has a partition table.

johnliu55tw:
I think the argument `auto-provision-enabled` is a bit redundant? If I want to disable auto-provision feature, I could run it without `auto-provision-filter` argument and there will be no auto-provision.

Maybe there's confusion between `auto-provision-enabled` and `force-format-disk`? For this PR, I see that NDM would always format disks that match the pattern without checking whether they already have a partition table (implies they might contain some data). If "always format" is acceptable behavior, I'd suggest only keeping `auto-provision-filter` argument.

weihanglo:
Sorry I misunderstood your previous comment. However, here comes a new question. Consider this scenario:

- `--auto-provision-filter=/dev/sda,/dev/sda2`
- `--force-format-disk=false`
- Got these devices: `/dev/sda`, `/dev/sda1`, `/dev/sda2`

What's the expected result? Should NDM provision `/dev/sda2` regardless that `/dev/sda` cannot be provisioned? Or should just reject all auto-provision settings?

pkg/filter/filter.go
```diff
@@ -12,38 +14,51 @@ type Filter struct {
...
 
-	vendorFilter := RegisterVendorFilter(vendorString)
-	pathFilter := RegisterPathFilter(pathString)
-	labelFilter := RegisterLabelFilter(labelString)
-	listFilter = append(listFilter, vendorFilter, pathFilter, labelFilter)
+	labels := strings.Split(labelString, ",")
+	labelFilter := RegisterLabelFilter(labels...)
 
-	return listFilter
+	return append([]*Filter{}, vendorFilter, pathFilter, labelFilter)
```
johnliu55tw, line 39: I think this could also be
```suggestion
	return []*Filter{vendorFilter, pathFilter, labelFilter)}
```

johnliu55tw:
> Sorry I misunderstood your previous comment. However, here comes a new question. Consider this scenario:
> 
>     * `--auto-provision-filter=/dev/sda,/dev/sda2`
> 
>     * `--force-format-disk=false`
> 
>     * Got these devices: `/dev/sda`, `/dev/sda1`, `/dev/sda2`
> 
> 
> What's the expected result? Should NDM provision `/dev/sda2` regardless that `/dev/sda` cannot be provisioned? Or should just reject all auto-provision settings?

If the auto-provision feature supports **only disks** (as we discussed with @guangbochen on Slack) , for your example **none** of the disks should be formatted because
- `/dev/sda` is a matched disk, but it has partition table and `--force-format-disk=false`, so ignored.
- `/dev/sda1` is a partition, ignored.
- `/dev/sda2` is a partition, ignored

But if we specify `--force-format-disk=true`, then `/dev/sda` should be formatted and added to LH.

Another thing is I just saw another option `AutoGPTGenerate`. Will this conflict with `--force-format-disk`?

weihanglo:
Oh I gave a wrong example. Sorry about that. There is still something obsolete. When provisioning a disk, NDM **provisions the only partition on a single partition disk** underneath the hood. I wonder in what scenario will NDM provision disks NDM with `--force-format-disk=false`. Maybe provisioning the first partition if exists?

Regard with `--auto-gpt-generate`, this is actually a helper for testing in virtualised environment. Since in that environment a "physical" disk may not have a WWN so we need a way to globally identify the disk and that's why we need a GPT on it. And yes, it may triggers GPT table partitioning twice (1st for get a GUID. 2nd for forceFormatted), but so far I see no harm for the entire flow.

guangbochen:
@johnliu55tw @weihanglo  it is correct that we don't need to deal with partition devices for now, 

Also if the `--force-format-disk` option is meaningless here we can consider removing it since all block devices should be force formatted automatically, and we can alarm the users on the docs and setting config of `the Harvester will auto force formatted those specified auto-provisioning devices, make sure they either backup or understand that clearly.`

pkg/filter/filter.go
weihanglo: Adrressed and rebased with ce09826.

main.go
```diff
@@ -119,6 +119,12 @@ func main() {
 			Usage:       "Enable auto GPT partition generating if a disk can not be globally identified",
 			Destination: &opt.AutoGPTGenerate,
 		},
+		&cli.StringFlag{
+			Name:        "auto-provision-filter",
+			EnvVars:     []string{"NDM_AUTO_PROVISION_FILTER"},
+			Usage:       "A string of comma-separated values that auto-provisions devices matching provided device path",
```
johnliu55tw, line 125: I think we could add some words here to indicate "we support shell Glob pattern"

johnliu55tw (approved):
LGTM. The help message is a good-to-have, not really needed for now.

main.go
guangbochen: +1 can we add a `support shell Glob pattern` with brackets on the usage quote, thanks.

weihanglo: Updated with 0de2ae9. Thanks folks

guangbochen (approved):
LGTM

guangbochen:
@weihanglo can u update the chart and bump the chart dep on the Harvester side, thanks.

PR #18, opened 2021-11-10
Exclude disks that contains longhorn in their busPath

weihanglo:
`ContainsIgnoredCase` is actually `MatchesIgnoredCase`, so we create
another `MatchesIgnoredCase` to do the real job.

This PR also includes a tiny fix for updating the right blockdevice.

### DIFF
--- a/pkg/filter/filter_test.go
+++ b/pkg/filter/filter_test.go
@@ -275,8 +275,8 @@ func Test_vendorFilter(t *testing.T) {
 		{
 			name: "longhorn bus path",
 			given: input{
-				disk:    &block.Disk{Vendor: "yourvendor", BusPath: "longhorn"},
-				vendors: []string{"longhorn"},
+				disk:    &block.Disk{BusPath: "ip-10.52.0.122:3260-iscsi-iqn.2019-10.io.longhorn:pvc-ab9af96e-60ef-400f-84f7-2f6eab68ab56-lun-1"},
+				vendors: []string{"LongHorN"},
 			},
 			expected: true,
 		},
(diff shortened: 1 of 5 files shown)


pkg/filter/filter_test.go
```diff
@@ -276,7 +276,7 @@ func Test_vendorFilter(t *testing.T) {
 			name: "longhorn bus path",
 			given: input{
 				disk:    &block.Disk{Vendor: "yourvendor", BusPath: "longhorn"},
-				vendors: []string{"longhorn"},
+				vendors: []string{"ip-10.52.0.122:3260-iscsi-iqn.2019-10.io.longhorn:pvc-ab9af96e-60ef-400f-84f7-2f6eab68ab56-lun-1"},
```
johnliu55tw, line 279: I think we should move this `ip-10.52.0.122:3260-iscsi-iqn.2019-10.io.longhorn:pvc-ab9af96e-60ef-400f-84f7-2f6eab68ab56-lun-1` string to the `BusPath` above as this is the BusPath of the disk? The `vendors` should be the strings provided by users.

pkg/util/util.go
```diff
@@ -42,11 +42,20 @@ func GetDiskPartitionPath(devicePath string, partitionNum int) string {
...
 		if strings.EqualFold(e, k) {
 			return true
 		}
 	}
 	return false
 }
+
+func ContainsIgnoredCase(s []string, k string) bool {
+	for _, e := range s {
+		if strings.Contains(e, k) {
```
johnliu55tw, line 56: - We need to swap the order of `k` and `e`, as the second argument of `Contains` is the substring to be checked.
- `Contains` is case-sensitive. Might need to tweak this function a bit to make it case-insensitive.

weihanglo: Fixed. Sorry for the rush.

johnliu55tw (approved):
LGTM

guangbochen (approved):

PR #19, opened 2021-11-29
New `pathTypeFilter` to exclude BIOS boot partition

weihanglo:
Related: harvester/harvester#1382, harvester/harvester#1241

## Solution

Introduce a new `pathTypeFilter`.

Partition like BIOS boot in GPT scheme should not be provisioned.
So do the parent disk of that partition. Thus, if any partition matches
excluded parttype, both the partition and its parent disk get excluded.

### DIFF
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -34,6 +34,7 @@ type Partition struct {
 	Label             string                  `json:"label"`
 	SizeBytes         uint64                  `json:"size_bytes"`
 	UUID              string                  `json:"uuid"` // This would be volume UUID on macOS, PartUUID on linux, empty on Windows
+	PartType          string                  `json:"part_type"`
 	DriveType         block.DriveType         `json:"drive_type"`
 	StorageController block.StorageController `json:"storage_controller"`
 	FileSystemInfo    FileSystemInfo          `json:"file_system_info"`
--- a/pkg/block/block_device.go
+++ b/pkg/block/block_device.go
@@ -273,6 +273,7 @@ func diskPartition(ctx *context.Context, paths *linuxpath.Paths, disk, fname str
 	du := GetDiskUUID(fname, string(PartUUID))
 	driveType, storageController := diskTypes(fname)
 	label := GetFileSystemLabel(fname)
+	partType := GetPartType(fname)
 	return &Partition{
 		Name:      fname,
 		Label:     label,
@@ -283,6 +284,7 @@ func diskPartition(ctx *context.Context, paths *linuxpath.Paths, disk, fname str
 			IsReadOnly: ro,
 		},
 		UUID:              du,
+		PartType:          partType,
 		DriveType:         driveType,
 		StorageController: storageController,
 	}
--- a/pkg/block/util.go
+++ b/pkg/block/util.go
@@ -9,21 +9,7 @@ import (
 )
 
 func GetParentDevName(devPath string) (string, error) {
-	if !strings.HasPrefix(devPath, "/dev") {
-		devPath = "/dev/" + devPath
-	}
-	args := []string{
-		"lsblk",
-		"-no",
-		"pkname",
-		devPath,
-	}
-	out, err := exec.Command(args[0], args[1:]...).Output()
-	if err != nil {
-		return "", fmt.Errorf("failed to get parent disk, %s", err.Error())
-	}
-
-	return strings.TrimSuffix(string(out), "\n"), nil
+	return lsblk(devPath, "pkname")
 }
 
 func HasPartitions(disk *Disk) bool {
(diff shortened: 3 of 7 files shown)


weihanglo:
Tested from my custom build. There is no blockdevice exist.

image

guangbochen (approved):
LGTM

pkg/block/util.go
```diff
@@ -9,45 +9,43 @@ import (
...
 	}
 	out, err := exec.Command(args[0], args[1:]...).Output()
 	if err != nil {
-		logrus.Warnf("failed to get filesystem label for device %s, %s", devPath, err.Error())
-		return ""
-	}
-	splited := strings.SplitN(string(out), "\n", 2)
-	if len(splited) > 0 {
-		return splited[0]
+		return "", fmt.Errorf("failed to get filesystem label for device %s, %s", devPath, err.Error())
```
johnliu55tw, line 47: Since this is a general function to execute `lsblk`, I think we also need to generalize this error message.

pkg/filter/part_type_filter.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+	}
+	return &Filter{
+		Name:       partTypeFilterName,
+		PartFilter: f,
+		DiskFilter: &diskPartTypeFilter{filter: f},
+	}
+}
+
+// Match returns true if partitionn matches given part types.
```
johnliu55tw, line 41: Typo
```suggestion
// Match returns true if partition matches given part types.
```

weihanglo: Nice catch. Updated.

pkg/block/util.go
weihanglo: Rebased and Updated. Thanks for the catch!

johnliu55tw (approved):
LGTM

pkg/block/util.go
```diff
@@ -9,45 +9,43 @@ import (
...
 
 func GetFileSystemLabel(devPath string) string {
+	result, err := lsblk(devPath, "label")
+	if err != nil {
+		logrus.Warnf(err.Error())
+	}
+	return result
+}
+
+func GetPartType(devPath string) string {
```
w13915984028, line 27: As "`lsblk`" may return "", err, but `GetPartType`, `GetFileSystemLabel `all only log the error, still return result ("")

the caller `diskPartition `will return a Partition with many areas blank to the upper caller ( the `controller`)

is it proper to skip those potential errors?

pkg/block/block_device.go
```diff
@@ -273,6 +273,7 @@ func diskPartition(ctx *context.Context, paths *linuxpath.Paths, disk, fname str
 	du := GetDiskUUID(fname, string(PartUUID))
```
w13915984028, line 273: Not relate to this commit, but a general doubt here:

```
func diskPartition(ctx *context.Context, paths *linuxpath.Paths, disk, fname string) *Partition {
	size := partitionSizeBytes(paths, disk, fname)
	mp, pt, ro := partitionInfo(ctx, paths, fname)
	...
```

`diskPartition `will skip all potential errors, e.g. following, return a partition with maybe blank/partial correct information, can it be checked and adjusted by the upper caller controller later?

```
431 func partitionSizeBytes(paths *linuxpath.Paths, disk string, part string) uint64 {
432         path := filepath.Join(paths.SysBlock, disk, part, "size")
433         contents, err := ioutil.ReadFile(path)
434         if err != nil {
435                 return 0
436         }
437         size, err := strconv.ParseUint(strings.TrimSpace(string(contents)), 10, 64)
438         if err != nil {
439                 return 0
440         }
441         return size * sectorSize
442 }
```

pkg/block/block.go
```diff
@@ -34,6 +34,7 @@ type Partition struct {
 	Label             string                  `json:"label"`
 	SizeBytes         uint64                  `json:"size_bytes"`
 	UUID              string                  `json:"uuid"` // This would be volume UUID on macOS, PartUUID on linux, empty on Windows
+	PartType          string                  `json:"part_type"`
```
w13915984028, line 37: Suggest to use "PartitionType" instead of "PartType", and all related parts.

pkg/block/util.go
weihanglo: In general, I don't think we can skip those errors. They should propagate up to the controller. Will address it in the following update.

pkg/block/block_device.go
weihanglo: The same. This need a code refactor. Thank for your advice!

guangbochen:
@w13915984028 would be nice if u can submit an enhancement PR of those flaws and assign @weihanglo to review, cheers.

PR #20, opened 2021-12-01
Fix reconcile enqueue

weihanglo:
Related: harvester/harvester#1599

Callback reenqueue: node-disk-manager migth be deployed earlier than Longhorn. Thus, if NDM cannot find longhorn node to provision, it should re-enqueue the callback after 15s.

## Test Plan

Refer to harvester/harvester#1599. 

You can 

1. deploy the first node, and then 
2. change the `--auto-provision-filter` arg for demonset/harvester-node-disk-manager to whatever you want to provision. 
3. Last, create a new join-cluster with disks matching that auto-provision pattern.


guangbochen (approved):
LGTM, seems like a reasonable fix, and would prefer to squeeze it into the rc1 release, what do u guys think?

johnliu55tw (approved):
LGTM

PR #21, opened 2021-12-07
Exclude disk whose driver type is neither HDD nor SSD

weihanglo:
Supposed this won't change over time, so I hard-coded this filter. 

Related: https://github.com/harvester/harvester/issues/1623#issuecomment-985108607


johnliu55tw (approved):

guangbochen (approved):
LGTM

PR #22, opened 2021-12-07
Two-phase disk unprovision 

weihanglo:
Related: harvester/harvester#1334

## Solution

Introduce Two-phase disk unprovision to prevent from data loss when user tries to remove a disk. (previously we remove disk immediately)

The new status indicates the provision phase of the device with
a Longhorn node. It's an enum type with three variants:

- `Provisioned`: Device is provisioned onto a Longhorn node as
  a disk. Longhorn replicas should be safe to schedule on it.
- `Unprovisioning`: The related disk of the device is under eviction
  from a Longhorn node. Once all replicas get evicted, the related
  disk would be removed from the node.
- `NotProvisioned`: The device is not provisioned at all. Might be
  possible to provision as a disk again.

The flow is simple:

1. User unprovisions the device/disk from the node by specifying
   `blockdevice.spec.fileSystem.provisioned = false`
2. NDM detects that a device/disk need to unprovision. To prevent
   data loss, NDM sets `evictionRequested = true`, 
   `allowScheduling = false` and disk tag `harvester-ndm-disk-remove` on the relevant disk on the longhorn
   node for rescheduling replicas to other disks. Meanwhile, the
   `blockdevice.status.provisionPhase` becomes `Unprovisioning`.
3. Once `scheduledReplica` of the disk is empty, NDM then safely
   removes the disk from the longhorn node.
   `blockdevice.status.provisionPhase` is then back to `NotProvisioned`.

## Test Plan

1. Create a 3-node cluster
2. Update blockdevice CRD and image of NDM daemonset
3. Attach some disk on a node, and then add disk onto the node in Harvester UI.
4. Create a VM and make sure your volume replicas is scheduled on the new disk (you can check it via Longhorn UI)
5. Go to node edit page and now remove the disk
6. Observe the blockdevice, you will see the phase transit from `Provisioned` to `Unprovisioning` to `NotProvisioned`. You can also check the Longhorn UI. The disk will first evict requested and not schedulable. Then it will be removed.

## TODO

- [ ] I guess the Havester UI need relevant updates?
- [ ] Update harvester/charts. Don't forget to update CRDs manifest.


pkg/controller/node/node.go
```diff
@@ -35,10 +37,48 @@ func Register(ctx context.Context, nodes ctllonghornv1.NodeController, bds ctldi
 		BlockDeviceCache: bds.Cache(),
 	}
 
+	nodes.OnChange(ctx, blockDeviceNodeHandlerName, c.OnNodeChange)
 	nodes.OnRemove(ctx, blockDeviceNodeHandlerName, c.OnNodeDelete)
 	return nil
 }
 
+func (c *Controller) OnNodeChange(key string, node *longhornv1.Node) (*longhornv1.Node, error) {
```
guangbochen, line 45: IIRC, longhorn reconciles its nodes in every 10s, so it is kind of resource-consuming to list and watch the node on change, can we consider moving the business logic to the BD side?

weihanglo: Sure. Off my head I can think of one solution: In blockdevice controller, `enqueueAfter` the blockdevice being unprovisioning to check whether or not the disk is removed. Does this sounds reasonable?

weihanglo: Updated. Thanks!

pkg/controller/blockdevice/controller.go
```diff
@@ -723,3 +757,14 @@ func (c *Controller) ApplyAutoProvisionFiltersForDisk(disk *block.Disk) bool {
 	}
 	return false
 }
+
+func (c *Controller) migrate(device *diskv1.BlockDevice) (*diskv1.BlockDevice, error) {
+	fs := device.Spec.FileSystem
+	phase := device.Status.ProvisionPhase
+	if phase == "" && fs.Provisioned {
```
johnliu55tw, line 764: Maybe we also have to migrate BlockDevice without the `ProvisionPhase` field?

pkg/controller/blockdevice/controller.go
```diff
@@ -537,33 +544,60 @@ func (c *Controller) removeDeviceFromNode(device *diskv1.BlockDevice) error {
 		return err
 	}
 
-	if _, ok := node.Spec.Disks[device.Name]; !ok {
-		logrus.Debugf("disk %s not found in disks of longhorn node %s/%s", device.Name, c.Namespace, c.NodeName)
-		msg := fmt.Sprintf("Disk not found in longhorn node `%s`", c.NodeName)
+	diskToRemove, ok := node.Spec.Disks[device.Name]
+	if !ok {
+		logrus.Debugf("disk %s not in disks of longhorn node %s/%s", device.Name, c.Namespace, c.NodeName)
```
johnliu55tw, line 549: Nit: I'd prefer to log error messages like this with `logrus.Errorf` so it's easier to trace the issue from Pod logs.

weihanglo: Updated. The disk would be umounted as well after unprovision finished.

johnliu55tw (approved):
LGTM

guangbochen (approved):
LGTM, but there are some known limitations of this backend implementation, we will need to add doc to ask the admin user to check more details via the Longhorn UI if any unexpected issue occurs:
- `UnProvisioning` will be pending forever if there is not sufficient disk space.
-  may consider allowing users to disable disk eviction on the next version.

cc @n313893254 we also need to show the status of`provisionPhase` of the blockDevice on the UI`.
image

PR #23, opened 2021-12-21
Phases for formatting, partitioning, mounting and provisioning

weihanglo:
## Problem

Related issue: harvester/harvester#1718

To prevent re-entry of critical sections in blockdevices reconcilation, introduce phases for those criticial operations, including

- Partitioning (on disk)
- Formatting (on partition)
- Mounting (on partition with a filesystem)
- Provisioning (on partition with a mounted filesystem)

Be aware of that when current phase is a `verb-ing`, blockdevice cannot perform any further operation. This is enforce in the reconcilation logic.

node-disk-manager-phase-transition

## Changes

- ❗ **Remove `NDM_AUTO_GPT_GENERATE` related env and flag**: If user wants to provision a disks without WWN or GPT partition, they need to create GPT partition first.
- ❗ **Remove `Spec.Filesystem.MountPoint`**: Use `Filesystem.Provisioned` instead to determine whether a provision is needed. The mount point now is an implementation detail and not configuralble.
- ❗ **Refactor the whole logic of blockdevice reconcilation**: Basically, we use a simple state machine pattern as the source of truths to determine which phase the blockdevice is. Also, we're not rely on `Status.DeviceStatus.FileSystem` anymore since it might be quite outdated, we now consult the system directly.
- 💅🏾 **Extract scan/filter logic out from blockdevice controller**: Tidy up the mess inside blockdevice controller. The new Scanner struct can be reuse in udev as well.

### KVM/libvirt env

> Note: You can attach disks via either commandline or virt-manager GUI, but you need to attach it with SCSI/IDE controller. That's because virtual disk in libvirt cannot be attached with a WWN.

1. Create a harvester cluster
2. Create a disk (cmd: `qemu-img create /path/to/disk.img 10G`)
3. Attach the disk to a node (cmd: `virsh attach-disk <vm-name> /path/to/disk.img sda --config --live --wwn 10000000c9b0513a`)
4. Head to dashboard **Hosts** > **Edit Config** > **Disks**. Then select any disk/partition to provision it.
5. Verify the provision succeeds.
6. Unprovision the disk via command (UI is not ready yet): `kubectl -n longnhorn edit bd`. Then find the resource of the blockdevice then set `spec.fileSystem.mountPoint = ""`.
7. Verify the disk is unprovisioned and unmounted.
8. Detach the disk (cmd: `virsh detach-disk harv-0 sdd --config --live`)
9. Verify the device is gone in both system and k8s blockdevice.

## TODO: UI change 

Note that multiple [conditions](https://github.com/harvester/node-disk-manager/pull/23/files#diff-adc97552573a07a88e490d13b406223be2bf84fc8cbda3f0bdee4b49ffae1471R10-R15) and [phase](https://github.com/harvester/node-disk-manager/pull/23/files#diff-adc97552573a07a88e490d13b406223be2bf84fc8cbda3f0bdee4b49ffae1471R230-R243) are added. Please read the changes if you want to display them.

#### `blockdevice.spec.fileSystem.mountPoint` has been removed.

The frontend does not need to provide mountpoint in order to provision a disk. 

- [ ] Setting `spec.fileSystem.provisioned = true` is sufficient to provision a disk
- [ ] Setting `spec.fileSystem.provisioned = true` is sufficient to trigger an unprovision.
- [ ] Remove mountpoint input field during editing. Users do not need to know that value when provisioned. This is the implementation detail.

#### Multiple new `blockdevice.status.provisionPhase` is added

When `spec.fileSystem.provisioned == true`, the system would start to provision given device. The frontend should:

- [ ] For blockdevice with partition type, prevent user from manipulate the blockdevice until the `status.provisionPhase == Provisioned`
- [ ] For blockdevice with disk type, once user specifies `spec.fileSystem.provisioned = true` then the user cannot do any further manipulation on it.
- [ ] **optional**: Display the current phase in the UI.
- [ ] Check `status.conditions.formatted` to determine if the device has been forced formatted or not. Do not let user format the device if it has been.

When `spec.fileSystem.provisioned == false`, that means the system would try to unprovision the device. The frontend should: (on applicable to device with provision type)

- [ ] For blockdevice with partition type, prevent user from manipulate the blockdevice until the `status.provisionPhase == Formatted`

## TODO: Others

- [ ] Need to update CRD manifest in relevant NDM chart and bump a version
- [ ] Support request burst control when multiple formatting happen simultaneously. #25
- [ ] Use filesystem UUID to determine which filesystem info to retrieve #24


weihanglo:
A glimpse of the new log messages:

```log
time="2021-12-23T07:10:35Z" level=info msg="Start GPT partition for e9a03081906692ecbff85a9c761f3028"
time="2021-12-23T07:10:35Z" level=info msg="Await GPT partition for e9a03081906692ecbff85a9c761f3028"
time="2021-12-23T07:10:35Z" level=info msg="make gpt partition of device /dev/sdc"
time="2021-12-23T07:10:35Z" level=info msg="Finished GPT partitioning for e9a03081906692ecbff85a9c761f3028"
time="2021-12-23T07:10:35Z" level=info msg="Add new block device dd40a9cb0e38604e52fff9484eaa2a9e with device: /dev/sdc1"
time="2021-12-23T07:10:35Z" level=info msg="Start formatting for dd40a9cb0e38604e52fff9484eaa2a9e"
time="2021-12-23T07:10:35Z" level=info msg="make ext4 format of the device /dev/sdc1"
time="2021-12-23T07:10:35Z" level=info msg="Await formating for dd40a9cb0e38604e52fff9484eaa2a9e"
time="2021-12-23T07:10:35Z" level=info msg="Update existing block device status dd40a9cb0e38604e52fff9484eaa2a9e with devPath: /dev/sdc1"
time="2021-12-23T07:10:35Z" level=info msg="Await formating for dd40a9cb0e38604e52fff9484eaa2a9e"
time="2021-12-23T07:10:35Z" level=info msg="Finished formatting dd40a9cb0e38604e52fff9484eaa2a9e"
time="2021-12-23T07:10:35Z" level=info msg="Start mounting dd40a9cb0e38604e52fff9484eaa2a9e on /var/lib/harvester/extra-disks/e9a03081906692ecbff85a9c761f3028"
time="2021-12-23T07:10:35Z" level=info msg="Await mounting dd40a9cb0e38604e52fff9484eaa2a9e on /var/lib/harvester/extra-disks/e9a03081906692ecbff85a9c761f3028"
time="2021-12-23T07:10:35Z" level=info msg="Finished mounting dd40a9cb0e38604e52fff9484eaa2a9e"
time="2021-12-23T07:10:35Z" level=info msg="Start provisioning dd40a9cb0e38604e52fff9484eaa2a9e on /var/lib/harvester/extra-disks/e9a03081906692ecbff85a9c761f3028"
time="2021-12-23T07:10:35Z" level=info msg="Finished provisioning dd40a9cb0e38604e52fff9484eaa2a9e on /var/lib/harvester/extra-disks/e9a03081906692ecbff85a9c761f3028"
```

pkg/controller/blockdevice/controller.go
```diff
@@ -682,76 +441,45 @@ func (c *Controller) OnBlockDeviceDelete(key string, device *diskv1.BlockDevice)
...
+			logrus.Debugf("disk %s not found in disks of longhorn node %s/%s", bd.Name, c.namespace, c.nodeName)
 			continue
 		}
-		existingMount := bd.Status.DeviceStatus.FileSystem.MountPoint
-		if existingMount != "" {
-			if err := disk.UmountDisk(existingMount); err != nil {
-				logrus.Warnf("cannot umount disk %s from mount point %s, err: %s", bd.Name, existingMount, err.Error())
+		filesystem := c.scanner.BlockInfo.GetFileSystemInfoByDevPath(bd.Spec.DevPath)
+		if filesystem.MountPoint != "" {
+			if err := disk.UmountDisk(filesystem.MountPoint); err != nil {
```
w13915984028, line 466: Is it OK to skip UmountDisk() failure?

pkg/controller/blockdevice/controller.go
```diff
@@ -4,509 +4,337 @@ import (
...
+	} else {
+		c.unmountFilesystem(bd, fs)
 	}
+}
 
-	// create a GUID partition table with a single partition only
-	if err := disk.MakeGPTPartition(device.Spec.DevPath); err != nil {
-		return device, err
+func (c *Controller) unmountFilesystem(bd *diskv1.BlockDevice, fs *block.FileSystemInfo) {
+	cmdErr := disk.UmountDisk(fs.MountPoint)
```
w13915984028, line 286: check cmdErr first

pkg/controller/blockdevice/controller.go
```diff
@@ -4,509 +4,337 @@ import (
...
-	// Update the single partition block device to start force formatting
-	toUpdate := bd.DeepCopy()
-	toUpdate.Spec.FileSystem.Provisioned = filesystem.Provisioned
-	toUpdate.Spec.FileSystem.MountPoint = filesystem.MountPoint
-	toUpdate.Spec.FileSystem.ForceFormatted = true
-	diskv1.DeviceFormatting.SetStatusBool(toUpdate, true)
-	diskv1.DeviceFormatting.Message(toUpdate, fmt.Sprintf("formatting disk partition %s with ext4 filesystem", toUpdate.Spec.DevPath))
-	if _, err := c.Blockdevices.Update(toUpdate); err != nil {
-		return device, err
+func (c *Controller) mountFilesystem(bd *diskv1.BlockDevice, fs *block.FileSystemInfo) {
```
w13915984028, line 307: check cmdErr first

pkg/controller/blockdevice/controller.go
```diff
@@ -4,509 +4,337 @@ import (
...
 
-// forceFormatPartition simply formats the partition to ext4 filesystem
-//
-// - umount the block device if it is mounted
-// - create ext4 filesystem formatting of the single partition
-func (c *Controller) forceFormatPartition(device *diskv1.BlockDevice) (*diskv1.BlockDevice, error) {
-	if device.Status.DeviceStatus.Details.DeviceType != diskv1.DeviceTypePart {
-		return device, fmt.Errorf("device %s is not a partition", device.Name)
+func (c *Controller) formatPartition(bd *diskv1.BlockDevice) {
+	cmdErr := disk.MakeExt4DiskFormatting(bd.Spec.DevPath)
```
w13915984028, line 253: When `MakeExt4DiskFormatting ` failure, still needs to update condition, e.g. set condition to init and update self, allow reconcile. Otherwise, the bd may stuck in "formatting" state

w13915984028:
The mechanism like "state machine" is used to handle hardware processing state, a couple of consideration points:
1. A bunch of conditions are used, it is the limitation of `condition.Cond`, which can only express true/false. The alternative way is one condition variable with different values expressing different states.
2. As a general rule, the state machine needs assistant of timer to avoid suspending in a state due to error handling or some other thing.  This needs clear design specification to express state switching and to design test cases.

pkg/block/blkid.go
```diff
@@ -19,7 +19,7 @@ func GetFileSystemType(part string) string {
 	}
 	out, err := exec.Command(args[0], args[1:]...).Output()
 	if err != nil {
-		logrus.Warnf("failed to read disk uuid of %s : %s\n", part, err.Error())
```
w13915984028, line 22: Hardware related issues are difficult to trouble shooting, maybe we need to keep it in `Warnf ` now.
1. Help debugging
2. When things like nvme controller partition is excluded from source code, the related error will disappear.
3. The function `GetFileSystemType `returns "" when error, the caller can only check "", not error.  Log error is the only clue for us.

weihanglo:
> The mechanism like "state machine" is used to handle hardware processing state, a couple of consideration points:
> 
>     1. A bunch of conditions are used, it is the limitation of `condition.Cond`, which can only express true/false. The alternative way is one condition variable with different values expressing different states.
> 
>     2. As a general rule, the state machine needs assistant of timer to avoid suspending in a state due to error handling or some other thing.  This needs clear design specification to express state switching and to design test cases.

Thank you! The mechanism really needs a more proper design. I have no experience of coining a state machine in Golang. Do you have any useful references at hand?

pkg/block/blkid.go
weihanglo: Reasonable. Would revert the changes.

pkg/controller/blockdevice/controller.go
weihanglo: I guess we can return error and let the controller retry.

weihanglo: I believe in reality a user should not unplug a parent disk without safely handle it's children partitions. That's my first assumption. Though we can return error and let the controller simply retries.

pkg/controller/blockdevice/controller.go
weihanglo: Correct. That's why it checks `cmdErr` later.

w13915984028:
> > The mechanism like "state machine" is used to handle hardware processing state, a couple of consideration points:
> > ```
> > 1. A bunch of conditions are used, it is the limitation of `condition.Cond`, which can only express true/false. The alternative way is one condition variable with different values expressing different states.
> > 
> > 2. As a general rule, the state machine needs assistant of timer to avoid suspending in a state due to error handling or some other thing.  This needs clear design specification to express state switching and to design test cases.
> > ```
> 
> Thank you! The mechanism really needs a more proper design. I have no experience of coining a state machine in Golang. Do you have any useful references at hand?

Two "state machine" can be referred:
`TCP state machine` (quite complex), `BGP state machine`, following article explains many details of BGP state machine:
http://net.cmed.us/Home/routing-theory/bgpv4/the-finite-state-machine

pkg/controller/blockdevice/controller.go
w13915984028: Normally speaking, the `cmdErr := disk.MakeExt4DiskFormatting(bd.Spec.DevPath)` cmdErr should be checked right before next call of `device, err := c.blockdeviceCache.Get(bd.Namespace, bd.Name)`

In current source code, if `device, err := c.blockdeviceCache.Get(bd.Namespace, bd.Name)` err happens, it will return, but the real return result of `"cmdErr"` is lost. The "cmdErr" is the first hand return result we need.  The expected `logrus.Infof("Failed to format %s: %v", device.Name, cmdErr)` will not be done.

Even though in error handling of `cmdErr  ` , it needs the result of `device, err := c.blockdeviceCache.Get(bd.Namespace, bd.Name)`, it may be nested in the error processing.

weihanglo:
This is not a formal state machine diagram but I guess it can describe the gist.

Each "verb-ing" state trigger a side effect action. the transition would not proceed until the action is complete or timeout, .




<details>

<summary>draw.io source code</summary>

```xml
<mxfile host="app.diagrams.net" modified="2022-01-03T06:12:33.380Z" agent="5.0 (Macintosh)" etag="YJhOSquO9jm3DtMzuSCb" version="16.1.2" type="device"><diagram name="Page-1" id="58cdce13-f638-feb5-8d6f-7d28b1aa9fa0">7R1rk5s28Nfcx2R4+vExd5dLO5N0rr02bfpNMbKtBiMHy/Y5v74SCBmEMLLNQ3chmUmMEAL2vavd5ca9Wz1/iMF6+QkHMLxxrOD5xr2/cRzbm7j0PzZySEemk0k6sIhRwCcdB57QD8gHLT66RQHcFCYSjEOC1sXBGY4iOCOFMRDHeF+cNsdh8a5rsIClgacZCMujf6OALPmobVnHE79AtFjyW098fuIrmH1bxHgb8ftFOILpmRXIluFTN0sQ4H1uyH1/497FGJP01+r5DoYMrBnE0useKs6KR45hRHQuiO7efHa+b/9wdneTH3MM5xH4/MZ20mV2INxyWNzDHZrBPw9reI823/izk0MGqs0erUJA39K95ZfCmMDnyoeyxatS6oF4BUl8oFP4BRlwON3Yfnq4PyLBG/Epyxz8vQmfCDjiF2LlIwToDw6EcwDilwBSAgFD+Lry5TmNgq/ZdKsMlBO4qASVWwSVUwaVO1WBym8AVL//CkOy37+Zb/1/D7vn3ReAoIp2/oq+RXgflSCWcAhki9mUbvZLRODTGszY2T2VKHRsSVYhPw3iGZcQjkcP5ygM73CI42Qpd07/zCw6viEx/gZzZ96/G92ORuxMxmls2gyv0Iz/DsFXGN4Kls2uTZiW3ghH5AGsUMgg/BnGAYgAH+bPQ984Pc7d1Ur+sMcO0SKiYzOKZRgL9qhHu19BSpWUMClSwqRMCbajoISMkZonBLeeZ+gyVJjDIqYpotbsPIyCJwIILGNbQLca20Xaeu24d7NLOPI9lRxQIN9tC/nZwieRH1Btyw9xTJZ4gSMQvj+O5pDIwHmc8xHjNcfsf5CQA0cG2BJcJCVKQu+YIcCQEILNhuGdDT6gUFBbJQmdia0N3sYzeIoh+EQC4gUkGiKUAegk9mMYAoJ2RXOleUbWUH4nGXlD35i8MlYOUEzVOcJskGKeMVl77G33zc72uEQCjxSniAEARYtBs3ep2Z3M/+hPtU+ryYFifaCGDqnBdXqnBl9HQfxkyl4ELGqVvWuUss8WzuHyAaBw4OqWudoZm8bW7kTB1aOQJMbPjv5csJ8bFEBGvfM5i8BRAtnOZhAmdPHuhj1vesXXOLsgG6HPlFtGsfKHxz9TUKd6peo6iSopuIlEejIOGVLQDITv+IkVCoJUAsEN+sEjNAz7a4wikoDVv73x79laVOhsON2UyDUjuRxt86HWqEYiGmescPq9Lg1Fyt09SP4TYoO5HplWuAwbtaI+Y5R6UT81StS7ZSvOLOTlFHpLqMv8mlrUZThuDnX80kcmY3Kenidx9FSK06avxK/Kx/VrFnLl2Hj6yqWFEkoS73M5cXmeBnEN0b9mwwOmRP+ccrigoOCtDDZ09PuWbXXdhnBOjkeyWcBgTC+bA87N2bw7yg2IAtexfoP78uWbNZy9pdQDnw4bAldv55jyzgOOV4AQZtFayWIWielTHs2L9GanrRVtqyN5r9dhc2S7caOyZSqc0jx5Oa1Zpi/PvezCDnF0lVmm9QyxQ+ypcfEDjqHedgocTUw6jWNSz5qwvYbMEhHL7MgscUenNZPaqZ3zCIim54pj7akErSDeksG5VYVEbJlWFN6tpVA8fmuKxzhBZaYmyrisXn7Zrcgv+nrgkJvAab1SvLlyFo1rSbSSrtioKBKizxxq6lvtaXvjo27UnkCRRBdnqz0Rza1aqGW151h2L6T1jMg/7PK3Pj/6kjtz/8xXTg4O/KDgdRf9u5e6xdO8lLtKi3k13nlqfLxfocT0Sb3lZJPfgjuYUDXFs2PNligMKCzpzyBJDh2MGJURUwzNiGTMvPOsis205jx7Kgt4sGEquaSWu7NIpyHc7ViOaQju26yY6mKyJ2/anVxoVsgLee2ZFWoUlGXJsU7g8bhzK1Ffs8UCvmSkKaoFfNXG+aSJ8LcaLCoFK8GgrWqB0UlY1VcLKAsrxtO2tJE9KmcZfKKihxscIArov48x2qFNRaphjphKIuU2+cvO8EkfUZRZBroAHVcgqRLGVTyZB7KKHt1TmRz8bn8woogWIcyRv4RTT5EdqKoAcdzi3UBIba8IEHjLJP/mSnFR4d+pPN6S5XnEv5Y9OeQh6bDXpeSroN6OswvH5Yy0KqKpjtEONNMhzdh+/0RjnBHclpdzGof1QQxLO3tpbFYYwx6XC8z+ilZCeQws3ybLO64U2Oib5cUTdMvh7YUru2F/VztPfWxY1sBYlWFWsgseY5z6DoNtYIRtIFKWe7QNfppd3GuFQ7YBryEcDAuCOipPc9AO5xKAfhXT2KwyJqGtCsbhOlMGg4HYuYHYv1MoXJ2+KpoS34QuwTKSN0lG8uvaMm2KdKSIpq2qd+62qskyo1SmyT21K1WDiOXWZz9YHdXOjCS6Ebs+526rlRYaSwu1na1ja8XJ2xNUIKDOksjqSNM9PuJoscRxpL/sIMTyfo+qnrdbIWarAuk/uRDTT+Gy+hFio0tTDksLyUkGbQsxRSeBToVYDFd4B/NybB7j1SDJzrfkbQ17bNypKMsk1wtx7/uXc7rJbIJr25ZzE6/GxtKVc6WFZKuvbTnn1si5vopdE28zA1aykBWh8KTQe62Vr01JQlvut+GVJaGoPuokm9dxB0l4niT0dCWh25EkHE2LRDW+1OKbSEUmI69rSdhL85djkYmjX2VC6Sk+pFeN/exYXMYOjtclR32Up3RG66Zt33jlQO3QQ/qCTiJn566a1kPaMy7u2l7Z2ViTrTP7or6cofG461W49FVBTRmXQxvhFrm7/zbCE43KCIOs1W4r07Tbt/hON5apaCd4ZvlzY/UCnipTpxToe8iVqA4ZXn3bDAZ0p56YUeJqtJTRSAzXbeIhoiKGGBq2ssS5SmwMeaEmSI3+2x0LrVpDNSdbQQ0k0yLJuHI3lv41zVTl0wztxLroxHE2+ZSa+fTeToxKwT7sEnVXHhEiNbQrj67dIuKjGu24DbNbpjo9k82xW1tq66GN6Mw7fDFdwBy7xp9ueYPGHqvCqQZsVQ99ma/Rb8b0ZbZ9Mxxv431rwYj1Oqp5IXeljjIjgmuOGvL06+cb//yHWs28mmbNjqVV4tteCmm+UId3QnztSaMNOVulDxP1nf1u+8YJLjN1k2A6Dd3UeBveK6u0jCvsNsVF0ijSajwDSFOnyMG5iz9wZElarnXlVLPLYFDe7/GiNaDiN7/E4Fxp6TPpA0uKyLOjqoBoL+/XMkOdvZhYkfYXCazGlZqm/GqsCawsUVtP+1WV1cukOHzqrZk9ME/qqqCQRG2ldClfwe3HwuLZ2+Ig2cF4KzK5K5K3L7GsJTlwoVRSTzzP0L5WBFV/FaXxvtCnLaO23fZtdaON1+S213TNPkGb1Xuk8leoXcUeaafyxQwz5/xv1DYjNk73WsoLjmo+bNxycX3JGr7UhSslAck77RWWy9mfd3LU92n3805uhs9hc7850s+StDTK9zrqOiPatst9rq42v+lhjJmffpweg/XyEw4gm/E/</diagram></mxfile>
```
</details>

weihanglo:
@johnliu55tw I think this is ready to review. Thanks!

pkg/controller/blockdevice/effects.go
```diff
@@ -0,0 +1,297 @@
...
+			logEffect(bd).Info("Finish partitioning")
+		} else {
+			logEffect(bd).Errorf("Failed to partition: %v", cmdErr.Error())
+		}
+
+		update := func(bd *diskv1.BlockDevice) *diskv1.BlockDevice {
+			if cmdErr != nil {
+				diskv1.ProvisionPhaseFailed.Set(bd)
+				setDevicePartitionedCondition(bd, corev1.ConditionFalse, cmdErr.Error())
+			} else {
```
w13915984028, line 47: Doubt: the "cmdErr" is not a param of "update" func, when will its value be confirmed, compile time or running time?

pkg/controller/blockdevice/effects.go
```diff
@@ -0,0 +1,297 @@
...
+			return err
+		}
+		if !errors.IsConflict(err) {
+			return emitError(err)
+		}
+		bd, err := e.BlockdeviceCache().Get(bd.Namespace, bd.Name)
+		if err != nil {
+			return emitError(err)
+		}
+		if _, err := e.Blockdevices().Update(update(bd.DeepCopy())); err != nil {
```
w13915984028, line 285: To solve the previous detected "IsConflict(err)? will it really work ?

And, the "update" function is called again, it uses "cmdErr" to verifiy, whats the value of "cmdErr" at this point?

pkg/controller/blockdevice/effects.go
johnliu55tw: `update` is a closure, so `cmdErr` inside `update` function is the `cmdErr` at line 36.

pkg/controller/blockdevice/effects.go
weihanglo: > To solve the previous detected "IsConflict(err)? will it really work ?

If the conflict occurs because of `resourceVersion` is too old, fetching the latest blockdevice resource might help.

> And, the "update" function is called again, it uses "cmdErr" to verifiy, whats the value of "cmdErr" at this point?

The same `cmdErr` captured at the lexical scope where the `update` closure created. That is, the value of `cmdErr` is same when the first call on `update`. the `cmdErr` itself is bound to `update` closure.

johnliu55tw (commented):
I am testing NDM under KVM with virtual storage devices with `WWN`, but there are still logs like this appeared every 30 secs:
```
time="2022-01-06T06:01:02Z" level=warning msg="failed to read disk uuid of /dev/sda : exit status 2\n"
time="2022-01-06T06:01:02Z" level=warning msg="failed to read disk uuid of /dev/sda : exit status 2\n"
time="2022-01-06T06:01:02Z" level=warning msg="failed to read disk uuid of /dev/sda : exit status 2\n"
```
These logs no longer appeared after I created GPT for the disk. Seems like our GUID relies on GPT's UUID? Can our system still works without GPT UUID? If so then I'd suggest remove the warning message.

pkg/controller/blockdevice/transitions.go
```diff
@@ -0,0 +1,228 @@
...
+		nodeName:         nodeName,
+		nodes:            nodes,
+		nodeCache:        nodes.Cache(),
+		blockdeviceCache: blockdeviceCache,
+		scanner:          scanner,
+	}
+}
+
+// next deduces the next phase from current blockdevice status
+func (p transitionTable) next(bd *diskv1.BlockDevice) (diskv1.BlockDeviceProvisionPhase, effect, error) {
```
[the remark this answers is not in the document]
weihanglo: My initial attempt was a type level design to determine in compile time, but turns out that Go's type system is not strong enough. I think your runtime solution is a good alternative.

pkg/controller/blockdevice/effects.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+
+		return updateDevice(e, bd, update)
+	}
+}
+
+// effectMountFilesystem mounts blockdevice onto its mountPoint.
+// Then update its ProvisionPhase to ProvisionPhaseMounted.
+func effectMountFilesystem(e effectController, bd *diskv1.BlockDevice) error {
+	mountPoint := bd.Spec.FileSystem.MountPoint
```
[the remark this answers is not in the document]
weihanglo: I assume a mount should be fast enough and shouldn't block the thread. But yes we can turn  those calling system/shell commands become a goroutine

pkg/controller/blockdevice/scanner.go
```diff
@@ -0,0 +1,261 @@
...
+
+	if oldBd, ok := oldBds[bd.Name]; ok {
+		newStatus := bd.Status.DeviceStatus
+		oldStatus := oldBd.Status.DeviceStatus
+		lastFormatted := oldStatus.FileSystem.LastFormattedAt
+
+		// Only disk hasn't yet been formatted can be auto-provisioned.
+		autoProvisioned = autoProvisioned && lastFormatted == nil && !diskv1.DeviceFormatted.IsTrue(oldBd)
+
+		if autoProvisioned || !reflect.DeepEqual(oldStatus, newStatus) || oldBd.Status.State != diskv1.BlockDeviceActive {
```
w13915984028, line 186: What will happen when "oldBd.Status.State != diskv1.BlockDeviceActive", say `v1beta1.BlockDeviceInactive`

```
func (u *Udev) deactivateBlockDevice will set

deviceCpy.Status.State = v1beta1.BlockDeviceInactive
```

pkg/controller/blockdevice/scanner.go
```diff
@@ -0,0 +1,261 @@
...
+		bd.Spec.FileSystem.MountPoint = fmt.Sprintf("/var/lib/harvester/extra-disks/%s", bd.Name)
+	}
+
+	if oldBd, ok := oldBds[bd.Name]; ok {
+		newStatus := bd.Status.DeviceStatus
+		oldStatus := oldBd.Status.DeviceStatus
+		lastFormatted := oldStatus.FileSystem.LastFormattedAt
+
+		// Only disk hasn't yet been formatted can be auto-provisioned.
+		autoProvisioned = autoProvisioned && lastFormatted == nil && !diskv1.DeviceFormatted.IsTrue(oldBd)
```
johnliu55tw, line 184: `autoProvisioned` will remain `true` even after the disk has been provisioned because `lastFormattedAt` is always `nil` and there's no condition `DeviceFormatted` for Disk type of BlockDevice.

weihanglo: Got it. That is to day, we don't need to deprecate `lastFormattedAt`. And checking the existence of  `lastFormattedAt` is sufficient to determine autoprovisioned or not.

johnliu55tw: If you feel like `lastFormattedAt` should be removed, then I think you could use other conditions and status added this time such as condition `Partitioned`.

pkg/controller/blockdevice/scanner.go
weihanglo: `BlockDeviceInactive` is flagged by udev.go, which udev thinks the device is about to remove (or perform some changes). At that circumstance NDM won't do a further change until the next scanning.

pkg/controller/blockdevice/effects.go
```diff
@@ -0,0 +1,297 @@
...
+	e.Blockdevices().EnqueueAfter(bd.Namespace, bd.Name, enqueueDelay)
+	return nil
+}
+
+// effectPrepareFormatPartition prepares required spec for child partition before start formating
+func effectPrepareFormatPartition(childBd *diskv1.BlockDevice) effect {
+	return func(e effectController, parentBd *diskv1.BlockDevice) error {
+		logEffect(parentBd).Infof("Start preparing format partition for its child %s", childBd.Name)
+		chilBdCpy := childBd.DeepCopy()
+		chilBdCpy.Spec.FileSystem.MountPoint = parentBd.Spec.FileSystem.MountPoint
```
johnliu55tw, line 73: Should the `MountPoint` of the child blockdevice be its name instead of using its parent's `MountPoint`?
```suggestion
		chilBdCpy.Spec.FileSystem.MountPoint = fmt.Sprintf("/var/lib/harvester/extra-disks/%s", chilBdCpy.Name)
```

pkg/controller/blockdevice/scanner.go
johnliu55tw: Or just use `Phase`:
```go
autoProvisioned = autoProvisioned && diskv1.ProvisionPhaseUnprovisioned.Matches(oldBd)
```

Although I'm still pondering whether we should use `Phase` here...

johnliu55tw:
I'm testing auto-provisioning 20 disks on KVM. 13 are added and the other 7 failed with error like this:
```
Failed to provision onto node harvmultidisk: Operation cannot be fulfilled on nodes.longhorn.io \"harvmultidisk\": the object has been modified; please apply your changes to the latest version and try again
```

I think it's because all those 20 disks are trying to modify the `node.longhorn.io` resource. A retry mechanism should solve this problem.

weihanglo:
All review comment are addressed. 

- Add a few basic unit tests for state transition.
- Remove `mountPoint` from spec in order to hide it as an implementation detail.
- Complete the retry and timeout mechanism

weihanglo:
New state machine diagram

node-disk-managet-phase-transition

<details>
<summary>draw.io XML</summary>

```xml
<mxfile host="app.diagrams.net" modified="2022-01-11T04:26:42.425Z" agent="5.0 (X11)" etag="gRtRtdG4QDSksM6krHdx" version="16.1.2" type="device"><diagram name="Page-1" id="58cdce13-f638-feb5-8d6f-7d28b1aa9fa0">7R1Zk5s4+rfMQ1ftPqQLJM7HJJ1kj0lVarNbO/NIbNpmg8ELuI/59SOMhK0DIZpDQHc/JEYWAuu7T93Aj4enL1lw3H9Nt2F8A4zt0w28uwHABNBF/5Ujz9WIa8JqYJdFWzzpMvA9+iPEgwYePUXbMKcmFmkaF9GRHtykSRJuCmosyLL0kZ52n8b0U4/BLuQGvm+CmB/9b7Qt9njUNIzLF38Lo90eP9qz8Rc/gs3PXZaeEvy8JE3C6ptDQJbBU/N9sE0fr4bgpxv4MUvTovp0ePoYxuW2kh2r7vvc8G39ylmYFCo3nJ5+7v7hbUBy+KfzW358/vL3u2/vTNfDb1c8k80It2hv8GWaFft0lyZB/Oky+uH8g8NyXQNdXeb8mqZHNGiiwf+FRfGMAR2cihQN7YtDjL9Fr5w9/1bef2uTy9/xcueLuyfq6plcPUVFdZtr48vfyZLo8+Wu8oLcVP3A8lc17hweytNTtgkl2+UQ5AyyXVhIJkID1iBGVBOmhxD9DnRnFsZBET3QrxJg9N3V8+pbv6UReklgYFLzoH/rO9Vdz4SGDHqV6lfgG68RglnLtbm1LHat6odya73PsuD5atqxnJA3v7ZLiIC8s2lIX83rOR868vmuLZ2PPlS/kFxdgfAydKavDrQGyGY/BPGJ0NrTEfEzREqIYSCiKl9xGxZ4SECXvwY/EOulaCmIo12CPm8QNocZGngIsyJCvO09/uIQbbcV2YZ59Efw47xeSRgYamhx+8ONfVeTSrlA+ERhKGa8+OYLU7smIhmD4QkBr//OuIWAMCEMjXf4oT1pBW04vazt0Uuk9/d5WDCAHwTUkNDRFai/ZelDlEdIQPCgzR+jQxwkPG/Niyz9GX5M4zQ7z4T++U8GKo6vNe69SROB5eLrx4sIBD4e219Jv/pGCSz+VWJLskM/sMvjal5GPQ7STwtihOdJUIQfyq3Kx4Cf6QEOfjfAidEbfdhGD+jjrvxIgbT6Fj3uagIH6At0S9J93EdF+P0YnGXOI1KtaMq+j+L4CvTbIPTuNyKkcDZe+ONeFSlkGNuMK0jwuBbNNAXwAwL4WRJsoQDXHUo2B6X/JEcClFIRzNID+i9Ok90+zcqRBOmuEqgYnaFy723CjRAqPzzbso1xoWJxUDFtW42sRgSLK5RdY+uUtW6orhq+VA+9aVcp5XKhVdM0PSysWlVNTAU9VM2e0PZkRBglu1UzwZq0NHJBg9vhN3LrTG5AmdwaFNpu2mpXC8r0OdTz5TZO6x0jWTker/r+cpTovss0azw5pyjNGsMG1O7jq95WDX3HeEaMKbBXPyX/P4WnUqnfnLLzHqHN2gd5OYI4RPnrslOyIgXLv/Wv/9x25j+prgU0+e8I879i/Vc+uWG5f454cvG+9PKigbODFY99jsrNwndsyYxNHOR5tKkG8RRzSiEyuNPPtGjL2Xmpw8+4BS2+w4H8ffW69Bs3iipmutfivZPNHkmquT7HCzmrEv1XvuX9ORywCjEHWrx3SMpZkALGUL47elXm/hFlnsfLvM9BFAv0lsVaMaxIo7UUof9gUoMGetZ8ZJqiSDMpgXaRb8sXaRgbFOwip6fw64k0WhUhk0Ia8DKscVeENurm9CA6U1cFpbZS68CnVONgp08RLoS+Vr8OjdGqmv1K2SD0/MnwuScb5PVUDovyfXAsP6L1omMeSjUV4/xXjqdJgZHG9PD1tZV+/uMQkNNv6vWqb0i2DRhV47GFxE7FOUVxx9EUHAPwSmaaHYJCFPbXExu2mWitye8ZtAWbBmW71hwbZiBkCxwrwkj09KFhkgKkBLsZ2wgNbIqsUzq+zGszgYIPa0MQI1CfzWD4S7EZBhJ6YplbfzN4JIWIkdYUOVXNz2lgQH2dZXyGnMtg3UBOLu45ACccNWmRjal7jX6xkgod4/Jnym4fSQ0Feq3xCcKL2tXQ4WnL1amEktekMgJIeiOScmH+nBfhQSItZxclkkvLkuXQXjRgeGjoinIBLyAnDRQ5Lq+2LM6vKYcC4KHgePWQvvwM4mBdLwMdn/ORRIZ2zqfV/CavyXO+heVBdaU002rndxNTneasKLAGslOud6kQf/KkKMsXYl1dNcKaqVJt25OavPLFxoo689HIU/IKakbak6tcSJtSJGDcN+5sMxk+Dr3CeJFnQYLB19UJDpOJo8xPbDh6/UirUNbUpYavRWqwSMgWzrLBP0s6f6xiQV6VFFQgHTtXIC1TIjgNmHKRCJYLGO/YIALBpBdFInkqgQB8vrpJgAFJKtQDXi8uIOXA85iCUjAMLtBJadNVkwJbq1DqENxYbEKUqmwjsmPmLlhBEcbXFTtgTZ8W0bNzv5oufNMrezsBVZNwmopzJsrBsbTaEK8icUsVZSBULaPtHY7uhzOQr6LlkGiFeVtytm6x4W+b4vL687ggD7SP4Oa9cZa1s83lEnXesDzBztUZXh2TuRSeJ3qchmwuNSu7BufSe3w0NKUi6zCFVdobfAgs4PUVoMphAhnnkyeAybQVpyS1f0nazcLygYit0Kbd+KpRcV9rixDg8XLyNRRPNpUC1I4qwwZ0uGmg6kkIqFV9+v4RqycNPsvo21U7pjODXnAzJjmrZnoQitTTiZ0OBm9ULC/pq6vOoltpAYajU0C+zNXTzfwfX7LVnKS94Mro24+nZ8D4rdlZf2gTg1EB2lrd6vWLigTcwjI1WvgqE/MW2R3TZmaQld8S+qawGDCmazMZfJ7Q1tdEzJLnuRm3DrToDm4DGQh0UoNF3z6ifWACDqrLzI9ucdrA2XFPqLdPWJd+/avgng2m/1QOF95xes4TOmbhQ5SecvRxG+U/zyNrY6qwteF82eSIps9BmCpbEjFZ6jDwl5dcsLT8IHVTRrlhUm/D9UXZr5BJlSENG0bNZq13RVISurCEpG5hmxk4A63lMYnZKQC1AqugAejtyk5WXkVX9m4u4Dlo229VJBMq203pL1Mp23yUZYUNz2Fr1r0P6Pz4gXwVjLPinTVdBYawmZqgBoc1qvIiOq/+Wqpz2nHDNRjcAMPgBr2oPRVmwPWncy/H4iJsZ/ahQsCnwaztuKoWTc2bn1UE1k7IU1CgasEvJgFdFAj12sCvojJDFWkgobz25MW+vUV64gwfs+KQaIWVGXJO7tCMXHslhrOYIwNGbf+p0EN+TgpZb8puiG4z2KnY/3Mw48AQdQGTeQSay4LwF5+vWEiNM00VQ8cwi9BvCLNSKSw9ffDOUWUGnQ4MFvSEthyR9gbHovs6yPaWFNSMhu2iWDX7Dvb1aTfQK9P922TlRMMBV4P5eWzeMIvyu8qXcwOc4FDaVNW/3KXxy134EKG3Qyw8Klbk8cO0JQ2ke6RJWt+2Wy7TSZlBgDFPau/Iq2ccJmmQpg1hEkd3lIRUqfCd9BcWnpLvOzTmFp6qCzvfksEmkZpgFKkJWW1scqnJq7lRXorBDlKz+eiMZcpM2JrR65l0b6qBzkukg2STVfwBm0/suU+zVwfYsgEdZE7bHuooTIttQTfZaZiQ14sruJas9angfS4zjky0CGmGmeqPTAC4+hDjBDJa1ccMNcclGuhsYWdKdVOEZ1CqC7W6j14UD+qYOD0BjSnrwVrjOMDi07TU1dXrpokr67aOiUCq2ljGQLoMoAsvWMfwiIrqco6xmoBkVQM0ekkW8rZF7WVdlYuIbTCu3UME+XMnvnz7Nxqo91+y+QtT/e3Zaf6m5lrNVWj+qmVYTdJvIhYnaJgkCSSthcPNQPW39UaOX6L6Ly0VbHgLQmvzF2DzxKoeJq7kZ6Wfr86EsFtNCNtyhjmhia4jmNKA0Bw0W4VUdhZB6QJXwfr6zrZkC7ASW3/jWXKs8Iqj1ssR2art+SytJTfA4T3rggq+zT6Kt9UGlXRNdaJdb9keaDortw5WA9t3BxHak5VsEmOe6s8XlltxhiiOXaIP5cZdQX1lPg0wP+atOZo5QaLu+pi3rbcHoODMeoZTr6lbuCOvuEE2lNPXhCL+ZeacjRHZMV9Av7wG1N3SaoX+42ldW45eS/Ulrq2ltQMbPNJmN7RQmMq1pbVhzasojB3cR6IXZaBnLg9lVqvQqfZKakpjnSraJTLHGSRaYS21XINgGzJqr6VeYPh5rrKA8Mj2bgjKBNzgtBm5ayh7zIpLdQ3l5ltev/kmPX+crqTQ0aP16BVHI6WkqWK68gHajS0oevoiPVCeKsqwXLXSKd6vaXNrWexaDd0GulKga7NiQk5RLfNHoihXq+nZRSkcug/H8HnZqhSl3nJ6IOHBdz5nVBiT9W000AC/ENu5Q7GmsSsxmQZT7+NCOTVxN0wjoFxePZ67J4cRamYLHcxPFVM/egyz1Kl1MdO2Zbg7EiqSIyVeH2fvaFwMiYqeOmcfR1myGIbsOGqcfTi04zPR4zD5S77Zh9vTOZXuIY1Ph/Cv6FP5dTnDECshiwsJYZqThYQs26KZARYYffvqAtGifeJF6DJL0+J6ehYc91/TbQmTT38C</diagram></mxfile>
```

</details>

manifests/crds/harvesterhci.io_blockdevices.yaml
```diff
@@ -65,16 +65,10 @@ spec:
...
                       overwrite the existing one
                     type: boolean
-                  mountPoint:
-                    description: a string with the partition's mount point, or ""
-                      if no mount point was discovered
-                    type: string
                   provisioned:
-                    description: a bool indicating whether the filesystem can be provisioned
-                      as a disk for the node to store data.
+                    description: a bool indicating whether the filesystem need to
```
w13915984028, line 69: `is needed`   or   `needs`

pkg/controller/blockdevice/transitions.go
```diff
@@ -0,0 +1,232 @@
...
+
+// transitionTable defines what phase the state manchine will move to
+type transitionTable struct {
+	namespace        string
+	nodeName         string
+	nodes            ctllonghornv1.NodeClient
+	nodeCache        ctllonghornv1.NodeCache
+	blockdevices     ctldiskv1.BlockDeviceClient
+	blockdeviceCache ctldiskv1.BlockDeviceCache
+	scanner          *Scanner
```
w13915984028, line 27: From the following call, what the `transitionTable ` needs is not a `scanner`, but a `block block.Info` ,  `block.New()` will return one.
Thus the `transitionTable `will be decoupled with `scanner`.

those calls:
```
108: part := p.scanner.BlockInfo.GetPartitionByDevPath(bd.Spec.DevPath, devPath)
131: 	filesystem := p.scanner.BlockInfo.GetFileSystemInfoByDevPath(bd.Spec.DevPath)
172: 		filesystem := p.scanner.BlockInfo.GetFileSystemInfoByDevPath(bd.Spec.DevPath)
207: 	filesystem := p.scanner.BlockInfo.GetFileSystemInfoByDevPath(bd.Spec.DevPath)
```

pkg/controller/blockdevice/effects.go
```diff
@@ -0,0 +1,423 @@
...
+					return
+				}
+				if _, err := e.Blockdevices().Update(update(newBd.DeepCopy())); err != nil {
+					emitError(err)
+					return
+				}
+			}
+		}
+	}()
+	go func() { f(doneCh) }()
```
johnliu55tw, line 422: I suppose this would be sufficient:
```suggestion
	go f(doneCh)
```

pkg/controller/blockdevice/effects.go
```diff
@@ -0,0 +1,423 @@
...
+				return err
+			}
+		}
+		return nil
+	}
+}
+
+// effectFormatPartition format given blockdevice to EXT4 filesystem, and then turns
+// ProvisionPhase to phaseFormatted.
+func effectFormatPartition(e effectController, bd *diskv1.BlockDevice) error {
```
w13915984028, line 97: Question to discuss:
When a lot of disks/partitions are added to the system at the same time, as per current processing, each blockdevice will go into its state machine. due to the fact that the "formatting" is quite slow,
1. many formatting go routines will run in parallel, should the controller has a counter (burst quota) to control the concurrency? too many formatting tasks in parallel wont help
2. when one formatting go routine is timed-out, it looks to not be terminated, the controller will re-enter that state and start a new go routine to do the formatting again. 
a. stop the previous go routine
b. wait until a formatting go routine finish (maybe more than 5 minutes)
c. keep current implementation,
what will happen when multi go routines are formatting the same disk/partition in parallel,   @johnliu55tw  your idea?

weihanglo: 1. It's great to have such feature, though I have no idea how many disks should be permitted to format together, and how normal a user formats lots disk at once. It not hard to implement but also non-trivial.

2. Did you spot any error? If I implemented it right, it should transit to failed and no one can re-enter to formatting phase for that blockdevice, unless user deletes that blockdevice to start over or modify the provisionPhase field manually.

johnliu55tw: 1. Formatting too many disks could cause some problems like CPU occupied by those formatting tasks, but I think it's a concern we could address later.
2. The essence of the state-machine-like design is to ensure the reconcile function is reentrant, so it should prevent creating multiple partitioning/formatting tasks of a single disk in parallel.

pkg/controller/blockdevice/effects.go
weihanglo: fixed 30893ee

pkg/controller/blockdevice/transitions.go
weihanglo: fixed with 87e65fe

manifests/crds/harvesterhci.io_blockdevices.yaml
weihanglo: fixed with 53b2bb3

pkg/controller/blockdevice/effects.go
```diff
@@ -0,0 +1,423 @@
...
+	go func() {
+		wait.JitterUntil(func() {
+			done, err := conditionFunc()
+			if done {
+				if err != nil {
+					logEffect(bd).Errorf("Ceased phase %s with err: %v", currentPhase, err.Error())
+				} else {
+					logEffect(bd).Infof("Finished phase %s", currentPhase)
+				}
+				stopCh <- struct{}{}
```
johnliu55tw, line 353: The document suggest closing the stopCh to stop JitterUntil [REF](https://github.com/kubernetes/apimachinery/blob/4427f8f31dfbac65d3a044d0168f84c51bfda440/pkg/util/wait/wait.go#L130)
```suggestion
				close(stopCh)
```

pkg/controller/blockdevice/effects.go
```diff
@@ -0,0 +1,423 @@
...
+			} else {
+				if err != nil {
+					logEffect(bd).Errorf("Failed during phase %s: %v", currentPhase, err.Error())
+				} else {
+					logEffect(bd).Infof("Continue polling in phase %s", currentPhase)
+				}
+			}
+		}, effectDefaultPollInterval, 1.0, false, stopCh)
+		doneCh <- struct{}{}
```
johnliu55tw, line 362: The document suggest closing the stopCh to stop JitterUntil [REF](https://github.com/kubernetes/apimachinery/blob/4427f8f31dfbac65d3a044d0168f84c51bfda440/pkg/util/wait/wait.go#L130)
```suggestion
		close(stopCh)
```

johnliu55tw (commented):
This logic appears multiple times:
```go
if !reflect.DeepEqual(resource.Status, resourceCpy.Status) {
	if _, err := e.ResourceController().Update(resourceCpy); err != nil {
		if errors.IsConflict(err) {
			// Hit optimistic lock. Fetch latest resource for next poll.
			newResource, err := e.ResourceCache().Get(resource.Namespace, resource.Name)
			if err != nil {
				// If a critical resource is missing, then abort this polling
				// because we cannot update its state anymore
				return errors.IsNotFound(err), err
			}
			resource = newResource
		}
		return false, err
	}
}
```

- Seems like this is trying to solve the resource update conflict issue, and `jitterPoll` would retry many times (for this kind of error) until timeout. However, `onCmdTimeout` also tries to do the same thing but it only retries **once**. I think once might not be enough.
- Enahncement: Can we extract this logic? Or merge this with `onCmdTimeout` to become something like `onCmdRetryTimeout`?

johnliu55tw:
Seems like NDM can't recover the disk from system rebooting. It just keeps trying to re-provision the disk onto Longhorn:
```
time="2022-01-12T08:08:02Z" level=info msg="Finished phase Unprovisioning" context=Effect device=f13387e74136b309cabbe51442ef02c3
time="2022-01-12T08:08:02Z" level=info msg="Start provisioning" context=Effect device=f13387e74136b309cabbe51442ef02c3
time="2022-01-12T08:08:02Z" level=info msg="Finish provisioning" context=Effect device=f13387e74136b309cabbe51442ef02c3
time="2022-01-12T08:08:02Z" level=info msg="Finished phase Provisioning" context=Effect device=f13387e74136b309cabbe51442ef02c3
time="2022-01-12T08:08:02Z" level=info msg="Start unprovisioning" context=Effect device=f13387e74136b309cabbe51442ef02c3
time="2022-01-12T08:08:02Z" level=info msg="Continue polling in phase Unprovisioning" context=Effect device=f13387e74136b309cabbe51442ef02c3
(repeat)
```

Because after a reboot, the partition would be unmounted, and seems like NDM doesn't know how to handle this in phase `Provisioned`.

pkg/controller/blockdevice/effects.go
weihanglo: fixed with ac4d8e1

pkg/controller/blockdevice/effects.go
```diff
@@ -0,0 +1,423 @@
...
+			if cmdErr != nil {
+				diskv1.ProvisionPhaseFailed.Set(bd)
+				setDevicePartitionedCondition(bd, corev1.ConditionFalse, cmdErr.Error())
+			} else {
+				diskv1.ProvisionPhasePartitioned.Set(bd)
+				setDevicePartitionedCondition(bd, corev1.ConditionTrue, "")
+			}
+			return bd
+		}
+	})
```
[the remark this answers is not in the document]
weihanglo: fixed with 38a1511

weihanglo:
Updated. Should be review-ready again. 🙇🏾 

- Add more uni tests for transitions
- `onCmdTimeout` now can retry if updating blockdevice fails
- Recover from "filesystem not mounted but is in phase Provisioned`

### ❗ Caution

There is a peril that NDM currently gets filesystem information from `blockInfo.GetFileSystemInfoByDevPath`. This device path is unstable after reboot. We need to seek the right partition by UUID. But I personally don't want to add more commits on this PR. Let's leave it to next one.

pkg/controller/blockdevice/controller.go
```diff
@@ -3,663 +3,121 @@ package blockdevice
...
-		nodeCpy := node.DeepCopy()
-		nodeCpy.Spec.Disks[device.Name] = diskToRemove
-		if _, err := c.Nodes.Update(nodeCpy); err != nil {
-			return err
+		var err error
+		deviceCpy := device.DeepCopy()
+		phase.Set(deviceCpy)
+		logrus.Debugf("[Transition] from %s to %s", device.Status.ProvisionPhase, deviceCpy.Status.ProvisionPhase)
+		device, err = c.blockdevices.Update(deviceCpy)
+		if err != nil {
```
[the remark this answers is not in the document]
weihanglo: Reasonable. Do you think it should be retry again?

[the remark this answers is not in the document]
weihanglo: No I think NDM shouldn't. It's already update and there must be an ongoing side effect taking over the transition.

johnliu55tw (approved):
The refactoring LGTM! Please help check why CI build is failing and update the description.

w13915984028 (approved):
LGTM

weihanglo:
Fixed the CI issue by removing unwanted `go:build`. Thanks everyone.

guangbochen (commented):
in general, the design change is looking good to me, but have we tested the upgrade case? seems the CRD change is handling its backward capability properly.

manifests/crds/harvesterhci.io_blockdevices.yaml
```diff
@@ -32,8 +32,8 @@ spec:
     - jsonPath: .spec.nodeName
       name: NodeName
       type: string
-    - jsonPath: .status.state
-      name: Status
+    - jsonPath: .status.provisionPhase
```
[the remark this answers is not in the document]
weihanglo: This is just for display in `kubectl get blockdevice`

pkg/apis/harvesterhci.io/v1beta1/types.go
```diff
@@ -6,10 +6,13 @@ import (
...
-	DeviceMounted    condition.Cond = "Mounted"
-	DeviceFormatting condition.Cond = "Formatting"
-	DiskAddedToNode  condition.Cond = "AddedToNode"
+const (
+	DevicePartitioned           condition.Cond = "Partitioned"
+	DeviceFormatted             condition.Cond = "Formatted"
+	DeviceMounted               condition.Cond = "Mounted"
+	DeviceProvisioned           condition.Cond = "Provisioned"
+	DeviceFailed                condition.Cond = "Failed"
+	DeviceAutoProvisionDetected condition.Cond = "AutoProvisionDetected"
```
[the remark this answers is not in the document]
weihanglo: It's like condition `AddedToNode` would linger there but cause no harm. NDM does not contain any logic to check that condition.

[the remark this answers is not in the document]
weihanglo: The caveat is that the UI does check `AddedToNode` and need to switch to condition `DeviceProvisioned`. It's ok if the UI and CRD are changed within the same version but we can also keep `AddedToNode` if needed.

weihanglo:
> in general, the design change is looking good to me, but have we tested the upgrade case? seems the CRD change is not handling its backward capability properly.

Haven't go through the full upgrade path, but has simulating an upgrade by 1) umounting a disk and update  manifest and then patch the image of the controller. The following scenarios are fine to me.

1. Attach a disk. And then upgrade: Some old conditions/statuses might left over. But as long as `spec` is in its initial state, the update won't affect anything. 
2. Attach a disk. Then provision it via UI. And Then upgrade: The disk should remain `provisioned`. If the old and new mountpoints mismatch, NDM will unprovision and unmount the disk and start over to provision.
3. Attach a disk with partitions (partitioned manually). Then provision the part. And then upgrade: Act like the first but with a partition

guangbochen (commented):
Also per discussion with @weihanglo, the current workflow of BDs will make it fall into a trap of forever pending when either the node is rebooted or the NDM pod is crashed.

As an enhancement, we may consider introducing a [work queue](https://pkg.go.dev/k8s.io/client-go@v0.23.2/util/workqueue#Type.Get) to process the disks concurrently, a single disk will only need to be processed when its status is either `*ing` or a newly added disk. If the disk is processed successfully it will update its status to either `success` or `fail` according to the runtime output.

weihanglo:
close in favor of #27

PR #25, opened 2022-01-18
[WIP] Limit max number of the concurrent format effect

weihanglo:
Related: https://github.com/harvester/harvester/issues/1831
Wait for #23 to be merged

This is not a fancy concurrent control mechanism. We are not using any concurrent primitives here but it works.

## Test plan

1. Create a disk and partition it to 6+ partitions (in terminal CLI)
2. `kubectl edit -n longhorn-system bd` and edit all partition with `spec.fileSystem = { partitioned: true, forceFormatted: true }` at once.
3. Observe the change of these blockdevice. Since the max concurrent count is 5 at this moment, you shall see some blockdevice are not started formatting in the first pass.
4. Finally, you shall see all blockdevice provisioned.


weihanglo:
Close since the design is totally different. Need to start over.

PR #26, opened 2022-01-19
Support raw block devices provision

weihanglo:
Related: https://github.com/harvester/harvester/issues/1839

Add support to provision raw blocks directly but maintain the ability to provision partitions.

- 58aacbae68295409b8b604ed1f0f0ea99b9efda6 **Remove auto-gpt-generate flag**: We do not provide this feature anymore. Devices are required to have WWN attached in order to uniquely identify by NDM.
- 3fb3e4bdd28789605d41de552c58273af23d4f3f **Remove the auto partition logic**: Raw block device can now be provisioned by NDM to Longhorn directly.
- e1444ae72990965921a17ba74b71ad9d46c918fc **Retrieve realtime filesystem info from OS**: NDM does not rely on CR status anymore to determine filesystem status. Instead, it queries the info from OS directly.
- 43ba0430076e7694bf444522c6b09b817f1c90e3 **Get real devpath from persistent device property**: In #24, NDM started retrieving filesystem info from filesystem UUID. This commit reverts a part of that logic. It's caller's responsibility to provide a correct devPath to `GetFileSystemInfoByDevPath`. In this commit, NDM therefore resolves the real devPath from persistent devPath before calling `GetFileSystemInfoByDevPath`. The `by-id` naming rules is defined in `/usr/lib/udev/rules.d/60-persistent-storage.rules`

#### Raw block device

1. Build NDM and patch the image of daemonset controller
2. Attach a raw block device to the node
3. `kubectl -n longhorn-system edit bd` and set `spec.fileSystem = { provisioned: true, forceFormatted: true, mountPoint: "/path/to/any/valid/mount/point" }`
4. The raw block device shall be schedulable as longhorn disk as a whole (without any partition) and is in `provisioned` phase
5. Set `spec.fileSystem.mountPoint = ""`, the device should be back to `Unprovisioned` phase

#### Partitions

1. Build NDM and patch the image of daemonset controller
2. Attach a raw block device to the node. Then partition the disk into multiple partitions, say, two.
3. `kubectl -n longhorn-system edit bd` and set `spec.fileSystem = { provisioned: true, forceFormatted: true, , mountPoint: "/path/to/any/valid/mount/point" } for all partitions`
4. All partitions shall be schedulable as longhorn disk as a whole (without any partition) and is in `provisioned` phase
5. Set `spec.fileSystem.mountPoint = ""` for all partitions. And then they should be back to `Unprovisioned` phase

PR #27, opened 2022-02-07
Fix race condition by updating resouce correctly

weihanglo:
Related issue: https://github.com/harvester/harvester/issues/1718

The root cause of race condition described in harvester/harvester#1718 can be resolved by https://github.com/harvester/harvester/issues/1718#issuecomment-1024114231. Specifically, it is a programming error that forgot to update blockdevice resource after formatting the disk.

This PR also fix the unnecessary update of DeviceMounted condition that causes a infinite update loop.

### Test plan

1. Prepare more than 10 extra disks for you harvester node.
2. Build NDM and update the version of harvester-node-disk-manager to this PR
3. Swipe all data on extra disks.
4. Enable auto-provision on NDM by adding an environment variable of the daemonset controller
  ```yaml
        - name: NDM_AUTO_PROVISION_FILTER
          value: /dev/sd*
  ```
5. Eventually, all disks matching the pattern should be partitioned, formatted and mounted successfully.
6. Check the log with the command `kubectl logs -n harvester-system ds/harvester-node-disk-manager -f`. After all disk being mounted, there should be no logs other than `Scan block devices of node`.

> Note that the partition and format timeout is not handled in this PR. That is a know caveat but does not directly lead to the race condition. We'll address it in the follow-ups.

### DIFF
--- a/pkg/controller/blockdevice/controller.go
+++ b/pkg/controller/blockdevice/controller.go
@@ -305,10 +305,13 @@ func (c *Controller) updateFileSystemStatus(device *diskv1.BlockDevice) error {
 	device.Status.DeviceStatus.FileSystem.MountPoint = filesystem.MountPoint
 
 	if filesystem.MountPoint != "" && filesystem.Type != "" {
-		err := isValidFileSystem(device.Spec.FileSystem, device.Status.DeviceStatus.FileSystem)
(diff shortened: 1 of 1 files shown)


w13915984028 (approved):
LGTM

weihanglo:
Close. Will reopen once #26 got merged

PR #29, opened 2022-02-11
Migrate base image to BCI image

bk201:
**Related issue**

https://github.com/harvester/harvester/issues/1868

### DIFF
--- a/package/Dockerfile
+++ b/package/Dockerfile
@@ -1,7 +1,11 @@
-FROM alpine
+FROM registry.suse.com/bci/bci-base:15.3
+
 # parted -> for `parted` command
-# util-linux -> for `lsblk` command
+# util-linux-systemd -> for `lsblk` command
 # e2fsprogs -> for `mkfs.ext4` command
-RUN apk update && apk add -u --no-cache parted util-linux e2fsprogs
+RUN zypper -n rm container-suseconnect && \
+    zypper -n install parted util-linux-systemd e2fsprogs iproute2 && \
+    zypper -n clean -a && rm -rf /tmp/* /var/tmp/* /usr/share/doc/packages/*
+
 COPY bin/node-disk-manager /usr/bin/
 CMD ["node-disk-manager"]


package/Dockerfile
```diff
@@ -1,7 +1,11 @@
-FROM alpine
+FROM registry.suse.com/suse/sle15:15.3
+
 # parted -> for `parted` command
-# util-linux -> for `lsblk` command
+# util-linux-systemd -> for `lsblk` command
 # e2fsprogs -> for `mkfs.ext4` command
-RUN apk update && apk add -u --no-cache parted util-linux e2fsprogs
+RUN zypper -n rm container-suseconnect && \
+    zypper -n install parted util-linux-systemd e2fsprogs && \
```
weihanglo, line 7: Need add `iproute2` package. (Dependency of go-iscsi-helper)
https://software.opensuse.org/package/iproute2

https://github.com/longhorn/go-iscsi-helper/blob/49a327fb024e4637ccc89ec6d73b3c550feec2d1/util/util.go#L94-L96

weihanglo (commented):
Note that the image size increases from 48MiB to 200MiB.

bk201:
> Note that the image size increases from 48MiB to 200MiB.

That's mostly caused by the dependency of package `util-linux-systemd`, in alpine world it's more simpler.

package/Dockerfile
```diff
@@ -1,7 +1,11 @@
-FROM alpine
+FROM registry.suse.com/bci/bci-base:15.3
```
bk201, line 1: Also switch from
```
FROM registry.suse.com/suse/sle15:15.3
```

to

```
FROM registry.suse.com/bci/bci-base:15.3
```

weihanglo (approved):
LGTM

bk201:
FWIW, using dive to inspect image layers:

original (alpine):
Screen Shot 2022-02-15 at 2 46 04 PM

new (BCI):

Screen Shot 2022-02-15 at 2 46 08 PM


The zstd-compressed tarball is 16M vs 63M.

guangbochen:
Wondering if `bci-base` is need for the NDM or using the `bci-minimal` is sufficient enough?

bk201:
@guangbochen  `zypper` is only available in the `bci-base` image. If we want a managed way to install dependent packages (like lsblk), the `bci-base` image is still preferred.

guangbochen (approved):
LGTM

PR #30, opened 2022-02-17
Apply label filter on disks

weihanglo:
Related: harvester/harvester#1924

## What's inside

After #26, raw block devices are enabled to be provisioned. As a result, label filter cannot filter out raw block since previously it excluded partitions only. This PR reuse the same logic of label filter on disks as well.

- Collect filesytem `LABEL` when a disk has a filesytem on it.
- Apply the same logic of label filter on disks as well.

## Test plan

Copy from the reproducible steps:

1. Install Harvester with default data disk set to a disk different from the OS disk, e.g. /dev/sda
1. Go to "Hosts" page -> Select the node -> "Edit Config" -> "Disks" page -> Click "Add Disk"
1. The default data disk should not in the drop-down list.

Another way to test:

1. Install a Harvester cluster
2. Add a disk. Format it as ext4 filesystem. 
3. Run `e2label /dev/your/disk HARV_MY_DISK`
4. The corresponding blockdevicee CR of your extra disk won't show in the result of `kubectl -n longhorn-system get bd`

### DIFF
--- a/pkg/filter/label_filter.go
+++ b/pkg/filter/label_filter.go
@@ -37,24 +37,15 @@ func RegisterLabelFilter(filters ...string) *Filter {
 
 // Match returns true if filesystem label matches the pattern
 func (f *partLabelFilter) Match(part *block.Partition) bool {
-	for _, pattern := range f.labels {
-		if pattern == "" || part.Label == "" {
-			return false
-		}
-		ok, err := filepath.Match(pattern, part.Label)
-		if err != nil {
-			logrus.Errorf("failed to perform filesystem label matching on disk %s for pattern %s: %s", part.Name, pattern, err.Error())
-			return false
-		}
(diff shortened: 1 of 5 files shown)


johnliu55tw (approved):
LGTM

pkg/filter/label_filter.go
```diff
@@ -37,24 +37,15 @@ func RegisterLabelFilter(filters ...string) *Filter {
...
-		}
-		if ok {
-			return true
-		}
-	}
-	return false
+	return matchDevLabel(part.Label, part.Name, f.labels)
 }
 
 // Match returns true if all partitions of the disk match.
```
bk201, line 43: nit: might be great to revise the comment as well.

bk201 (approved):
LGTM

PR #32, opened 2022-02-18
Enforce that only `OnBlockDeviceChange` can update CR

weihanglo:
## What does this try to resolve?

Related issue: harvester/harvester#1718

The problem of the original issue is that during disk formatting in a `OnBlockDeviceChange` call, `ScanBlockDevicesOnNode` periodically updates info of all block device CRs. When the disk formatting is finished and then tries to update the CR, it might hit the resource conflict error because `ScanBlockDevicesOnNode` have just updated the resources. That's where the race condition occur.

The basic gist of this pull request is to ensure that **only `OnBlockDeviceChange` can update** existing blockdevices. `ScanBlockDevicesOnNode` and uevent action handler only handle **create** and **delete** block device CRs.

Other changes in this PR:

- Extract scanner to its own file (CodeFactor forces me to do that...)
- Remove unused `status.state`.
- Lower the level of some noisy loggings.

## Test Plan

> ⚠️ This is a heuristic test plan since real world race condition is hard to reproduce. If you find any better alternative, feel free to comment.
> 
> This test is better to perform under QEMU/libvirt environment.

1. Prepare more than 10 extra disks for you harvester node.
2. Build NDM and update the version of harvester-node-disk-manager to this PR
3. Swipe all data on extra disks.
4. Enable auto-provision on NDM by adding an environment variable of the daemonset controller
    ```yaml
          - name: NDM_AUTO_PROVISION_FILTER
            value: /dev/sd*
    ```
5. Eventually, all disks matching the pattern should be partitioned, formatted and mounted successfully.
6. Check the log with the command `kubectl logs -n harvester-system ds/harvester-node-disk-manager -f`. After all disk being mounted, there should be no logs emitted.
7. Navigate to longhorn dashboard to see if each disk is successfully added and scheduled.

### DIFF
--- a/main.go
+++ b/main.go
@@ -195,23 +195,28 @@ func run(opt *option.Option) error {
 	autoProvisionFilters := filter.SetAutoProvisionFilters(opt.AutoProvisionFilter)
 
 	start := func(ctx context.Context) {
-		if err := blockdevicev1.Register(
-			ctx, lhs.Longhorn().V1beta1().Node(),
-			disks.Harvesterhci().V1beta1().BlockDevice(),
+		bds := disks.Harvesterhci().V1beta1().BlockDevice()
+		nodes := lhs.Longhorn().V1beta1().Node()
+		scanner := blockdevicev1.NewScanner(
+			opt.NodeName,
+			opt.Namespace,
+			bds,
 			block,
-			opt,
 			excludeFilters,
 			autoProvisionFilters,
+		)
+		if err := blockdevicev1.Register(
+			ctx,
+			nodes,
+			bds,
+			block,
+			opt,
+			scanner,
 		); err != nil {
 			logrus.Fatalf("failed to register block device controller, %s", err.Error())
 		}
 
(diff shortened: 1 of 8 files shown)


johnliu55tw (approved):
Overall LGTM! Just an unused field could be removed.

bk201:
LGTM and the function work great. There is a CRD change, could you create a PR in the charts repo in order to test upgrade? Thanks!

weihanglo:
@bk201 Thanks for the reminder. Here it is: https://github.com/harvester/charts/pull/69

weihanglo:
@bk201 After compatibility consideration, I'll keep `status.state` as-is. Just removed corresponding commit and rebased. Thanks.

weihanglo:
Hi, @guangbochen . I guess it's ready to merge. Would you like to take a look?

guangbochen (approved):
Scrawl approved as we already got 2 reviewers' LGTM, thanks.

PR #33, opened 2022-02-25
Don't expose `blockdevice.spec.fileSystem.mountPoint` to user

weihanglo:
## What does this try to solve

Related: harvester/harvester#1936

Don't expose `mountpoint` to user. This should be an implementation details.

Compatibility notice: If there is a previous non-empty mount point exists, node-disk-manager would continue using that value.

#### Case 1: provision and unprovision back-and-forth

1. Prepare a harvester cluster (single node is sufficient)
1. Prepare an extra disk and add it to the host via Harvester Dashboard.
1. Expect the disk
   - is mounted (check with CLI `lsblk`, `mount`, etc.), and 
   - volumes can be scheduled onto it. (check its existence in longhorn dashboard)
1. Remove the disk from the host via Harvester Dashboard
1. Expect the disk 
   - is unmounted (check with CLI `lsblk`, `mount`, etc.), and
   - volumes cannot be scheduled onto it. (check its existence in longhorn dashboard)
1. Provision it again and see if the disk become schedulable again.
1. Provision it again and see if the disk is removed again.

#### Case 2: provision then reboot

1. Prepare a harvester cluster (single node is sufficient)
1. Prepare an extra disk and add it to the host via Harvester Dashboard.
1. Expect the disk
   - is mounted (check with CLI `lsblk`, `mount`, etc.), and 
   - volumes can be scheduled onto it. (check its existence in longhorn dashboard)
1. Reboot the host and wait for the harvester cluster ready
1. Expect the disk
   - is mounted (check with CLI `lsblk`, `mount`, etc.), and 
   - volumes can be scheduled onto it. (check its existence in longhorn dashboard)
1. Remove the disk from the host via Harvester Dashboard
1. Reboot the host and wait for the harvester cluster ready
1. Expect the disk 
   - is unmounted (check with CLI `lsblk`, `mount`, etc.), and
   - volumes cannot be scheduled onto it. (check its existence in longhorn dashboard)

#### Case 3: compatible with non-empty mount point

See: https://github.com/harvester/node-disk-manager/pull/33#issuecomment-1062669133

### DIFF
--- a/manifests/crds/harvesterhci.io_blockdevices.yaml
+++ b/manifests/crds/harvesterhci.io_blockdevices.yaml
@@ -32,8 +32,8 @@ spec:
     - jsonPath: .spec.nodeName
       name: NodeName
       type: string
-    - jsonPath: .status.state
-      name: Status
+    - jsonPath: .status.provisionPhase
+      name: ProvisionPhase
       type: string
     - jsonPath: .metadata.creationTimestamp
       name: Age
@@ -66,8 +66,9 @@ spec:
                       overwrite the existing one
                     type: boolean
                   mountPoint:
-                    description: a string with the partition's mount point, or ""
-                      if no mount point was discovered
+                    description: 'DEPRECATED: no longer use and has no effect. a string
+                      with the partition''s mount point, or "" if no mount point was
+                      discovered'
                     type: string
                   provisioned:
                     description: a bool indicating whether the filesystem can be provisioned
--- a/pkg/controller/blockdevice/controller.go
+++ b/pkg/controller/blockdevice/controller.go
@@ -107,9 +107,7 @@ func (c *Controller) OnBlockDeviceChange(key string, device *diskv1.BlockDevice)
 		return device, err
 	}
 
-	// mount device by path, and skip mount partitioned device
-	needUpdateMount := filesystem != nil && filesystem.MountPoint != deviceCpy.Spec.FileSystem.MountPoint
-	if needUpdateMount {
+	if needUpdateMountPoint(deviceCpy, filesystem) {
 		err := c.updateDeviceMount(deviceCpy, devPath, filesystem)
 		if err != nil {
 			err := fmt.Errorf("failed to update device mount %s: %s", device.Name, err.Error())
@@ -123,11 +121,11 @@ func (c *Controller) OnBlockDeviceChange(key string, device *diskv1.BlockDevice)
 		return device, err
 	}
 
-	needProvision := deviceCpy.Spec.FileSystem.MountPoint != "" && deviceCpy.Spec.FileSystem.Provisioned
+	needProvision := deviceCpy.Spec.FileSystem.Provisioned
 	switch {
 	case needProvision && device.Status.ProvisionPhase == diskv1.ProvisionPhaseUnprovisioned:
 		if err := c.provisionDeviceToNode(deviceCpy, devPath); err != nil {
-			err := fmt.Errorf("failed to provision device %s to node %s on path %s: %w", device.Name, c.NodeName, device.Spec.FileSystem.MountPoint, err)
+			err := fmt.Errorf("failed to provision device %s to node %s: %w", device.Name, c.NodeName, err)
 			logrus.Error(err)
 			diskv1.DiskAddedToNode.SetError(deviceCpy, "", err)
 			diskv1.DiskAddedToNode.SetStatusBool(deviceCpy, false)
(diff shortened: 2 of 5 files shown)


weihanglo:
@johnliu55tw @bk201 this is ready to review. Thanks! 🙏🏾

manifests/crds/harvesterhci.io_blockdevices.yaml
```diff
@@ -32,8 +32,8 @@ spec:
     - jsonPath: .spec.nodeName
       name: NodeName
       type: string
-    - jsonPath: .status.state
-      name: Status
```
bk201, line 36: Like the previous PR, do we want to keep this field?

weihanglo: This is only for display purpose. The field itself is not removed

guangbochen (approved):
LGTM

weihanglo:
Just a rebase onto master. Sorry for the noise

bk201:
I saw a 1.0.0 blockdevice becomes `Unschedulable`, here is how to reproduce it:
- Create a 1.0.0 cluster.
- Add some disks with WWN to nodes.
- Provision a disk from GUI. Assume `/dev/sda`. `/dev/sda1` will be added to Longhorn.
- Pause the `harvester` ManagedChart.
- Apply the new CRD manifest: `kubectl apply -f manifests/crds/harvesterhci.io_blockdevices.yaml`
- Modify NDM daemon set to use the PR image, pods are restarted.
- Existing block device turns into `Unschedulable`.
Screen Shot 2022-03-09 at 4 24 20 PM

Note: I didn't upgrade Longhorn or harvester. Simply bump the NDM in a 1.0.0 cluster.

bk201:
The log on node4 might be helpful:
```
time="2022-03-09T08:19:29Z" level=info msg="Node Disk Manager c65eb4e (c65eb4e) is starting"
time="2022-03-09T08:19:29Z" level=info msg="Starting node disk manager controller"
time="2022-03-09T08:19:29Z" level=info msg="register exclude filters"
time="2022-03-09T08:19:29Z" level=info msg="register auto provision filters"
time="2022-03-09T08:19:29Z" level=info msg="Starting harvesterhci.io/v1beta1, Kind=BlockDevice controller"
time="2022-03-09T08:19:29Z" level=info msg="Unmount device 07dcae6ffdeb1c81bbf2ad6e69960276 from path /var/lib/harvester/extra-disks/e3ec6a8bc743fc53e48fddfde588995e"
time="2022-03-09T08:19:29Z" level=info msg="Mount deivce 07dcae6ffdeb1c81bbf2ad6e69960276 to /var/lib/harvester/extra-disks/07dcae6ffdeb1c81bbf2ad6e69960276"
time="2022-03-09T08:19:30Z" level=info msg="Start monitoring udev processed events"
time="2022-03-09T08:19:30Z" level=info msg="Starting longhorn.io/v1beta1, Kind=Node controller"
```

FrankYang0529 (approved):
Test and it work. LGTM.

johnliu55tw (approved):
LGTM

weihanglo:
> Add some disks with WWN to nodes.

@bk201 Did the underlying mount path change? In the previous version, user can specify the mount path. Therefore, if you patch the NDM daemon directly without rebooting, the mount points mismatch and it will become unschedulable.

I expect that the NDM upgrade will happen right after the reboot so the mountpoint will be updated.

weihanglo:
This PR will be postponed due to the upgrade path has not yet been handled.

weihanglo:
@bk201 I've resolved the compatibility issue. Please help review it again. Thanks!

pkg/controller/blockdevice/controller.go
```diff
@@ -326,9 +325,6 @@ func (c *Controller) unprovisionDeviceFromNode(device *diskv1.BlockDevice) error
 
 	updateProvisionPhaseUnprovisioned := func() {
 		msg := fmt.Sprintf("Disk not in longhorn node `%s`", c.NodeName)
-		// To prevent user from mistaking unprovisioning from umount, NDM umount
-		// for the device as well while unprovisioning it.
-		device.Spec.FileSystem.MountPoint = ""
```
johnliu55tw, line 331: If this line got removed, the `Spec.FileSystem.MountPoint` of a blockdevice will not be removed (assuming the blockdevice was provisioned in an older version of Harvester, so it has `Spec.FileSystem.MountPoint`). Is this the desired behavior?

weihanglo: Hmm… I orginally thought that we shouldn't specify `spec.filesytem.mounPoint` any more, but a cleanup here seems reasonable as you already tried to unprovision. Which behaviour do you think is better?

weihanglo: Oh. Just remember why I removed it. It's an anti-pattern modifying `spec` during controller reconcilation. I now lean towards leave mountpoint untouched.

johnliu55tw (approved):

bk201 (approved):
LGTM

PR #34, opened 2022-03-02
Support resolve real dev path from UUID and PTUUID

weihanglo:
Related: harvester/harvester#1874

### What does it resolve?

Starting from #26, NDM resolves device links to actual block device files. However, it only takes WWN into consideration. Since the GUID of blockdevice CR depends on WWN, UUID, and PTUUID (in this fablback order), NDM also needs a fallback chain to resolve to actual block device files.

#### Plug-unplug disks

1. Prepare a harvester cluster (single node is sufficient)
2. Prepare two disks
    - Both without WWN
    - Format one of the disk
    - Partition the other disk
3. Plug-in both disks and expect two blockdevice CR to show up without errors
4. Unplug both disks and wait for 30s, expect two blockdevice CR to be removed.
5. Plug-in both disk again.
    - Expect both disks to show up without errors
    - Expect both disks to get the same `Name`s as the last time they plugged in.

#### Swapping disks

1. Prepare a harvester cluster (single node is sufficient)
2. Prepare two disks and format both of them.
3. Hotplug both disks and add them to the host via Harvester Dashboard
4. Shutdown the host.
5. Swap the address and slot of the two disks in order to make their devpath swapped
    - For libvirt environment, you can swap `<address>` and `<target>` in the XML of the disk.
6. Reboot the host
7. Navigate to the "Host" page, both disks should be **healthy** and **scheduled**.

### DIFF
--- a/pkg/controller/blockdevice/controller.go
+++ b/pkg/controller/blockdevice/controller.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"time"
 
+	ghwutil "github.com/jaypipes/ghw/pkg/util"
 	lhtypes "github.com/longhorn/longhorn-manager/types"
 	lhutil "github.com/longhorn/longhorn-manager/util"
 	"github.com/sirupsen/logrus"
@@ -445,14 +446,26 @@ func (c *Controller) OnBlockDeviceDelete(key string, device *diskv1.BlockDevice)
 func resolvePersistentDevPath(device *diskv1.BlockDevice) (string, error) {
 	switch device.Status.DeviceStatus.Details.DeviceType {
 	case diskv1.DeviceTypeDisk:
-		wwn := device.Status.DeviceStatus.Details.WWN
-		if wwn == "" {
-			return "", fmt.Errorf("WWN not found on device %s", device.Name)
+		valueExists := func(value string) bool {
+			return value != "" && value != ghwutil.UNKNOWN
 		}
-		if device.Status.DeviceStatus.Details.StorageController == string(diskv1.StorageControllerNVMe) {
-			return filepath.EvalSymlinks("/dev/disk/by-id/nvme-" + wwn)
+		// Disk naming priority.
+		// #1 WWN
+		// #2 filesystem UUID (UUID)
+		// #3 partition table UUID (PTUUID)
+		if wwn := device.Status.DeviceStatus.Details.WWN; valueExists(wwn) {
+			if device.Status.DeviceStatus.Details.StorageController == string(diskv1.StorageControllerNVMe) {
+				return filepath.EvalSymlinks("/dev/disk/by-id/nvme-" + wwn)
+			}
+			return filepath.EvalSymlinks("/dev/disk/by-id/wwn-" + wwn)
 		}
-		return filepath.EvalSymlinks("/dev/disk/by-id/wwn-" + wwn)
+		if fsUUID := device.Status.DeviceStatus.Details.UUID; valueExists(fsUUID) {
+			return filepath.EvalSymlinks("/dev/disk/by-uuid/" + fsUUID)
+		}
+		if ptUUID := device.Status.DeviceStatus.Details.PtUUID; valueExists(ptUUID) {
+			return block.GetDevPathByPTUUID(ptUUID)
+		}
+		return "", fmt.Errorf("WWN/UUID/PTUUID not found on device %s", device.Name)
 	case diskv1.DeviceTypePart:
 		partUUID := device.Status.DeviceStatus.Details.PartUUID
 		if partUUID == "" {
--- a/pkg/controller/blockdevice/controller.go
+++ b/pkg/controller/blockdevice/controller.go
@@ -2,7 +2,9 @@ package blockdevice
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"reflect"
 	"time"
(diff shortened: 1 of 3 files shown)


weihanglo:
Please ignore CodeFactor CI failure. The complex issue has been addressed in #32

pkg/controller/blockdevice/controller.go
```diff
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"time"
 
+	ghwutil "github.com/jaypipes/ghw/pkg/util"
```
w13915984028, line 10: What is the consideration to import ghwutil only for a string "unknown" ?  or define it in local pkg/util..

https://github.com/jaypipes/ghw/blob/main/pkg/util/util.go
const (
	UNKNOWN = "unknown"
)

pkg/controller/blockdevice/controller.go
```diff
@@ -600,14 +601,26 @@ func (c *Controller) ApplyAutoProvisionFiltersForDisk(disk *block.Disk) bool {
...
+		valueExists := func(value string) bool {
+			return value != "" && value != ghwutil.UNKNOWN
+		}
+		// Disk naming priority.
+		// #1 WWN
+		// #2 filesystem UUID (UUID)
+		// #3 partition table UUID (PTUUID)
+		if wwn := device.Status.DeviceStatus.Details.WWN; valueExists(wwn) {
+			if device.Status.DeviceStatus.Details.StorageController == string(diskv1.StorageControllerNVMe) {
+				return filepath.EvalSymlinks("/dev/disk/by-id/nvme-" + wwn)
```
w13915984028, line 613: define `"/dev/disk/by-id/nvme-"`, `"/dev/disk/by-uuid/"` as local const ?

pkg/controller/blockdevice/controller.go
weihanglo: It's because the "unknown" here is filled by that package.

pkg/controller/blockdevice/controller.go
weihanglo: Hmm... I am not sure about this. Any benefit in defining them as constants than literal strings?

w13915984028: From my perspective, define them as const brings benefits of:
1. highlight them in the beginning of source code
2. could be reused
3. in case of change, one instead of many

weihanglo: Thanks. These points are all valid. However, I don't see any other place reusing these paths. (And it shouldn't because we want all devPath to be resolved by this function). Unless there is another reason such as memory efficiency, I think for this specific case we don't need to declare separately. Declaring constants on the top of the source code might also make code reading back-and-forth and pile up the cognitive load.

weihanglo:
Update `mkfs.ext4 -U <uuid>` with previous UUID value and rebased onto master.

pkg/controller/blockdevice/controller.go
```diff
@@ -213,7 +218,19 @@ func (c *Controller) forceFormat(device *diskv1.BlockDevice, devPath string, fil
 
 	// make ext4 filesystem format of the partition disk
 	logrus.Debugf("make ext4 filesystem format of device %s", device.Name)
-	if err := disk.MakeExt4DiskFormatting(devPath); err != nil {
+	// Reuse UUID if possible to make the filesystem UUID more stable
+	var uuid string
+	if !valueExists(device.Status.DeviceStatus.Details.WWN) {
+		uuid = device.Status.DeviceStatus.Details.UUID
```
w13915984028, line 224: if valueExists(device.Status.DeviceStatus.Details.WWN), should uuid be WWN ?

pkg/controller/blockdevice/controller.go
```diff
@@ -445,14 +462,23 @@ func (c *Controller) OnBlockDeviceDelete(key string, device *diskv1.BlockDevice)
...
+		if fsUUID := device.Status.DeviceStatus.Details.UUID; valueExists(fsUUID) {
+			return filepath.EvalSymlinks("/dev/disk/by-uuid/" + fsUUID)
 		}
-		if device.Status.DeviceStatus.Details.StorageController == string(diskv1.StorageControllerNVMe) {
-			return filepath.EvalSymlinks("/dev/disk/by-id/nvme-" + wwn)
+		if ptUUID := device.Status.DeviceStatus.Details.PtUUID; valueExists(ptUUID) {
+			return block.GetDevPathByPTUUID(ptUUID)
 		}
-		return filepath.EvalSymlinks("/dev/disk/by-id/wwn-" + wwn)
+		return "", fmt.Errorf("WWN/UUID/PTUUID not found on device %s", device.Name)
```
w13915984028, line 481: `was not found`

pkg/controller/blockdevice/controller.go
weihanglo: Node. We need filesystem UUID to be stable because if there is no WWN in the device, NDM will choose UUID/PTUUID as its unique identifier for the blockdevice custom resource.

weihanglo:
The latest commit addresses the problem: When UUID is reused in `mkfs.ext4` but the corresponding blockdevice CR haven't/failed to update. Then the next call on `OnBlockDeviceChange` cannot determine the real UUID to query disk info.

johnliu55tw (approved):
LGTM

pkg/controller/blockdevice/controller.go
w13915984028: Well, if L223 `// Reuse UUID if possible to make the filesystem UUID more stable` has a further comment of `"When WWN is not there, then reuse ..."`, I will not be doubting L225, which takes care of UUI, PtUUID, but skips WWN.

weihanglo: That's good advice. I'll update the comment. Thanks!

weihanglo: Updated with https://github.com/harvester/node-disk-manager/pull/34/commits/9ad6ccfbde62658e5acf6a6009242f3e0d79b09d
Thanks!

w13915984028 (approved):
LGTM, thanks.

PR #36, opened 2022-04-27
Update devpath if changed

weihanglo:
When reboot, device path might change but controller cannot detect
it during the reconciliation. We explicitly check and update the value
when scanner startup.

related: https://github.com/harvester/harvester/issues/2149#issuecomment-1110488173

### DIFF
--- a/pkg/controller/blockdevice/scanner.go
+++ b/pkg/controller/blockdevice/scanner.go
@@ -132,6 +132,9 @@ func (s *Scanner) scanBlockDevicesOnNode() error {
 			if s.NeedsAutoProvision(oldBd, autoProvisioned) {
 				logrus.Debugf("Enqueue block device %s for auto-provisioning", bd.Name)
 				s.Blockdevices.Enqueue(s.Namespace, bd.Name)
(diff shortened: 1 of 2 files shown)


pkg/controller/blockdevice/scanner.go
```diff
@@ -132,6 +132,9 @@ func (s *Scanner) scanBlockDevicesOnNode() error {
 			if s.NeedsAutoProvision(oldBd, autoProvisioned) {
 				logrus.Debugf("Enqueue block device %s for auto-provisioning", bd.Name)
 				s.Blockdevices.Enqueue(s.Namespace, bd.Name)
+			} else if isDevPathChanged(oldBd, bd) {
+				logrus.Debugf("Enqueue block device %s for device patch change", bd.Name)
```
bk201, line 136:
```suggestion
				logrus.Debugf("Enqueue block device %s for device path change", bd.Name)
```

bk201 (changes requested):

bk201 (approved):
LGTM

johnliu55tw (approved):
Verified dev path is updated. LGTM

PR #40, opened 2022-06-24
controller: make parameter really work with cli module

Vicente-Cheng:
- `DefaulText` is real text. That shows on `--help`. `Value` does really for the default value.
       Actually, if the `DefaultText` is nil, that would get the `Value` for it.

Signed-off-by: Vicente Cheng <vicente.cheng@suse.com>

### DIFF
--- a/main.go
+++ b/main.go
@@ -48,12 +48,14 @@ func main() {
 		},
 		&cli.StringFlag{
 			Name:        "namespace",
+			Value:       "longhorn-system",
 			DefaultText: "longhorn-system",
 			EnvVars:     []string{"LONGHORN_NAMESPACE"},
 			Destination: &opt.Namespace,
 		},
 		&cli.IntFlag{
 			Name:        "threadiness",
+			Value:       2,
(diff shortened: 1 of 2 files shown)


Vicente-Cheng:
That is the related fix for https://github.com/harvester/harvester/issues/1831

bk201:
To be complete, we also need to add the `NDM_MAX_CONCURRENT_OPS` environment variable in the chart template: https://github.com/harvester/charts/blob/master/charts/harvester-node-disk-manager/templates/daemonset.yaml

Vicente-Cheng:
> To be complete, we also need to add the `NDM_MAX_CONCURRENT_OPS` environment variable in the chart template: https://github.com/harvester/charts/blob/master/charts/harvester-node-disk-manager/templates/daemonset.yaml

Thanks for the reminder, done.

guangbochen (approved):
LGTM

bk201 (approved):
LGTM, work as expect.

PR #42, opened 2022-09-28
fix the disk tag missing when NDM respawn

Vicente-Cheng:
related issue: https://github.com/harvester/harvester/issues/2832

### DIFF
--- a/main.go
+++ b/main.go
@@ -165,6 +165,9 @@ func initLogs(opt *option.Option) {
 	}
 	logrus.SetOutput(os.Stdout)
 	logrus.Infof("Node Disk Manager %s is starting", version.FriendlyVersion())
+	logrus.Infof("Notable parameters are following:")
+	logrus.Infof("Namespace: %s, ConcurrentOps: %d, RescanInterval: %d",
+		opt.Namespace, opt.MaxConcurrentOps, opt.RescanInterval)
 	if opt.Debug {
 		logrus.SetLevel(logrus.DebugLevel)
 		logrus.Debugf("Loglevel set to [%v]", logrus.DebugLevel)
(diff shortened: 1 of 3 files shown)


futuretea (approved):
LGTM

main.go
```diff
@@ -165,6 +165,9 @@ func initLogs(opt *option.Option) {
 	}
 	logrus.SetOutput(os.Stdout)
 	logrus.Infof("Node Disk Manager %s is starting", version.FriendlyVersion())
+	logrus.Infof("Notable parameters are following:")
+	logrus.Infof("NameSpace: %s, ConcurrentOps: %d, RescanInterval: %d",
```
guangbochen, line 169:
```suggestion
	logrus.Infof("Namespace: %s, ConcurrentOps: %d, RescanInterval: %d",
```

guangbochen (approved):
small twist and everything else is LGTM.

main.go
Vicente-Cheng: done, thanks!

Vicente-Cheng:
hi all, fixed again with the auto-provision case.
Please help to review!

pkg/controller/blockdevice/scanner.go
```diff
@@ -128,6 +128,11 @@ func (s *Scanner) scanBlockDevicesOnNode() error {
 	for _, device := range allDevices {
 		bd := device.bd
 		autoProvisioned := device.AutoProvisioned
+		/* do not need to enqueue again with ProvisionPhaseProvisioned disk */
+		if bd.Status.ProvisionPhase == diskv1.ProvisionPhaseProvisioned {
+			logrus.Infof("Skip the provisioned device: %s", bd.Name)
+			continue
+		}
```
bk201, line 135: Not sure if this change caused issues to lines 140 ~143? (The `else if isDevPathChanged(oldBd, bd) {` clause)
IIRC, it's used to fix device path changes (e.g. swap disks when shutdown and boot)

Vicente-Cheng: yap, that's a problem. We need this to work well when the host reboots.
Fixed this case by respecting disk tags from LH.

please review again!

bk201 (approved):
Thanks!

futuretea (approved):
LGTM

PR #44, opened 2022-10-18
Update go mod

chrisho:
Related module:
- golang.org/x/net
- golang.org/x/text

Related self-test case:
https://harvester.github.io/tests/manual/_incoming/extra-disk-using-raw-disk/
https://harvester.github.io/tests/manual/hosts/1623-add-disk-to-host/

PR #45, opened 2022-10-19
Bump golang.org/x/net and golang.org/x/text

Vicente-Cheng:
upgrade following dependency packages
golang.org/x/net -> 0.0.0-20220906165146-f3363e06e74c
golang.org/x/text -> 0.3.8


Vicente-Cheng:
close due to #44

PR #52, opened 2023-02-17
ci: add validations on action runners

Vicente-Cheng:
We could do that on the action runner. That does not take much time.


bk201 (approved):
LGTM!

guangbochen (approved):

PR #60, opened 2023-11-30
controller: add disk tags to BlockDeviceSpec

Vicente-Cheng:
**Problem:**
We would like to add tags when we add extra disks.
(We need to add tags after we provision the disk )

**Solution:**
Add `Tags` to BlockDeviceSpec, then the controller can add it when provisioning

**Related Issue:**
https://github.com/harvester/harvester/issues/2717

**Test plan:**
TBD


Vicente-Cheng:
Need new Longhorn deployment script, depends on https://github.com/harvester/vagrant-rancherd/pull/8

pkg/controller/blockdevice/controller.go
```diff
@@ -190,6 +191,28 @@ func (c *Controller) OnBlockDeviceChange(_ string, device *diskv1.BlockDevice) (
...
+					logrus.Debugf("Prepare to update deivce %s because %v is not in %v", deviceCpy.Name, tag, deviceCpy.Status.Tags)
+					if err := c.provisionDeviceToNode(deviceCpy); err != nil {
+						err := fmt.Errorf("failed to update tags %v with device %s to node %s: %w", deviceCpy.Spec.Tags, device.Name, c.NodeName, err)
+						logrus.Error(err)
+						c.Blockdevices.EnqueueAfter(c.Namespace, device.Name, jitterEnqueueDelay())
+					}
+					break // only provision once
+				}
+			}
+		}
```
tserong, line 215: The same inner block is duplicated here ("Prepare to update device... c.Blockdevices.EnqueueAfter"). Is there any way to simplify this? Maybe using `slices.Compare` instead of the `if len != { ... } else { }` construct?

I checked https://pkg.go.dev/golang.org/x/exp/slices#Compare and it requires the slices to be ordered, so I guess it'd need to be something like this:

```
slices.Sort(deviceCpy.Spec.Tags)
slices.Sort(deviceCpy.Status.Tags)
if slices.Compare(deviceCpy.Spec.Tags, slices.Sort(deviceCpy.Status.Tags) != 0 {
    /* do the update here */
}
```

The kink with my suggestion is that this does mean potentially reordering the tags. Would that cause any problems?

pkg/controller/blockdevice/controller.go
```diff
@@ -371,22 +394,35 @@ func (c *Controller) provisionDeviceToNode(device *diskv1.BlockDevice) error {
...
-		logrus.Debugf("Previous disk tags on LH: %+v, we should respect it.", disk.Tags)
-		diskSpec.Tags = disk.Tags
-		needUpdated = reflect.DeepEqual(disk, diskSpec)
+		if disk.Tags != nil {
+			/* we should respect the disk Tags from LH */
+			logrus.Debugf("Previous disk tags on LH: %+v, we should respect it.", disk.Tags)
+			diskSpec.Tags = removeDuplicated(disk.Tags, device.Spec.Tags)
+			if len(toRemovedTags) > 0 {
+				logrus.Debugf("Prepare to do final handling with toRemovedTags: %+v", toRemovedTags)
+				diskSpec.Tags = removeUnNeeded(disk.Tags, toRemovedTags)
```
tserong, line 415: Should this be `diskSpec.Tags = removeUnNeeded(diskSpec.Tags, toRemovedTags)`? If I'm reading it correctly, the current form, which checks `disk.Tags`, will effectively overwrite whatever changes were made by the earlier call to `removeDuplicated`, i.e. there may still be duplicate tags at this point.

(I guess it could also be collapsed to something like `diskSpec.tags = removeUnNeeded(removeDuplicated(disk.Tags, device.Spec.Tags), toRemovedTags)`, but that's probably harder to read, and you'd also lose the logging)

tserong (commented):
Just to make sure I understand, this is meant to take tags defined via either harvester _or_ longhorn, and add them to the block device spec, and then to the block device status. If so, it looks like it should generally do the right thing, I just have a couple of comments.

Commit 25afa250af039e8575ea81f9b5e8c5dca1b42e20 ("controller: add disk tags to BlockDeviceSpec") does two things - it updates go.mod etc. with new/updated dependencies, and it also makes those changes to the controller. I'd suggest splitting that into two separate commits, one for the dependency update, and one for the main code change.

manifests/crds/harvesterhci.io_blockdevices.yaml
```diff
@@ -84,6 +84,11 @@ spec:
               nodeName:
                 description: name of the node to which the block device is attached
                 type: string
+              tags:
+                description: a string with for device tag for provisioner, e.g. "default,small,ssd"
```
tserong, line 88: Typo (extra "for") - should be "a string with device tag for provisioner..."

Vicente-Cheng: Nice catch!  thanks

pkg/controller/blockdevice/controller.go
Vicente-Cheng: nice catch, yes, it should be `diskSpec.Tags = removeUnNeeded(diskSpec.Tags, toRemovedTags)`

I would like to avoid collapsing these two functions because that makes read more easy and logging.

pkg/controller/blockdevice/controller.go
Vicente-Cheng: Good suggestion! Update with your suggestion!
BTW, I thought that would be ok for the order of the tags.

tserong (approved):
LGTM

tserong (changes requested):
Oops, no, wait a sec - I read the code but didn't check the tests, one of which has failed: https://github.com/harvester/node-disk-manager/actions/runs/7112001697/job/19361613496?pr=60

Vicente-Cheng:
> Oops, no, wait a sec - I read the code but didn't check the tests, one of which has failed: https://github.com/harvester/node-disk-manager/actions/runs/7112001697/job/19361613496?pr=60

This failed because `longhorn.io/v1beta2/nodes` updated too frequently (means we will provision failed and queue again), CI does not know it. I am working on it.

Vicente-Cheng:
Depends on https://github.com/harvester/go-common/pull/8
I moved some general functions to the above repo. Once it is merged, I will update the PR again.

BTW, CI is also fixed now.

Vicente-Cheng:
Hi @tserong, Please help recheck this PR.
Thanks!

tserong (approved):
LGTM!

pkg/controller/node/node.go
```diff
@@ -30,15 +35,67 @@ func Register(ctx context.Context, nodes ctllonghornv1.NodeController, bds ctldi
...
+		}
+
+		logrus.Debugf("Prepare to checking block device %s", name)
+		bd, err := c.BlockDevices.Get(c.namespace, name, metav1.GetOptions{})
+		if err != nil {
+			logrus.Warnf("Get block device %s failed: %v", name, err)
+			return node, err
+		}
+
+		if bd.Spec.Tags == nil || len(bd.Spec.Tags) == 0 {
```
bk201, line 73: Tags don't sync back if a BD never had tags before.

Vicente-Cheng: Update with the latest changes. Do not change to the previous design.

Now the `status.Tags` of `blockdevice` should be the same as `nodes.longhorn.io`
`Spec.Tags` of `blockdevice` contains the tags set up by harvester.

bk201 (approved):
LGTM, thanks!

Vicente-Cheng:
@mergifyio backport v0.6.x

PR #64, opened 2023-12-07
drone: remove binary release

Vicente-Cheng:
We do not release binary after this PR.
Users can get it from the image.


Vicente-Cheng:
@mergifyio backport v0.5.x

PR #66, opened 2023-12-18
process: do not enable profiler by default

Vicente-Cheng:
**Problem:**
NDM created with the `hostNetwork: true`. This parameter will expose the whole port on this pod to the host.
So, we need to prevent listening to unnecessary ports on NDM pods.

**Solution:**
we can disable profiler by default. NDM does not use it generally.

**Related Issue:**
https://github.com/harvester/harvester/issues/4212

**Test plan:**
make sure the host does not listen :6060
```
harvester-node-0:~ # ss -tnal |grep 6060
LISTEN 0      4096               *:6060             *:*
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -69,9 +69,8 @@ func main() {
 		},
 		&cli.StringFlag{
 			Name:        "profile-listen-address",
-			Value:       "0.0.0.0:6060",
-			DefaultText: "0.0.0.0:6060",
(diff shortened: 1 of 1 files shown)


bk201 (approved):
LGTM

tserong (approved):

Vicente-Cheng:
@mergifyio backport v0.5.x

PR #76, opened 2024-01-24
uevent: should respawn the uevent monitor

Vicente-Cheng:
**Problem:**
The udev monitor will not work with any errors.

**Solution:**
We monitor the error and try to respawn it.

**Related Issue:**
https://github.com/harvester/harvester/issues/4925

**Test plan:**
Because the original error is hard to reproduce, I added another inject error here.

1. set the parameter `inject-udev-monitor-error` as true.
2. try to add a disk and provision it should work

### DIFF
--- a/pkg/udev/uevent.go
+++ b/pkg/udev/uevent.go
@@ -19,28 +19,45 @@ import (
 )
 
 type Udev struct {
-	namespace string
-	nodeName  string
-	startOnce sync.Once
-	scanner   *blockdevice.Scanner
+	namespace   string
+	nodeName    string
+	startOnce   sync.Once
+	scanner     *blockdevice.Scanner
+	injectError bool
 }
 
 func NewUdev(opt *option.Option, scanner *blockdevice.Scanner) *Udev {
 	return &Udev{
-		startOnce: sync.Once{},
-		namespace: opt.Namespace,
-		nodeName:  opt.NodeName,
-		scanner:   scanner,
+		startOnce:   sync.Once{},
+		namespace:   opt.Namespace,
+		nodeName:    opt.NodeName,
+		scanner:     scanner,
+		injectError: opt.InjectUdevMonitorError,
 	}
 }
 
 func (u *Udev) Monitor(ctx context.Context) {
-	u.startOnce.Do(func() {
-		u.monitor(ctx)
-	})
+	// we need to respawn the monitor with any error.
+	// because any error will break the monitor loop.
+	errChan := make(chan error)
+	go u.spawnMonitor(ctx, errChan)
+
 }
 
-func (u *Udev) monitor(ctx context.Context) {
+func (u *Udev) spawnMonitor(ctx context.Context, errChan chan error) {
+	go u.monitor(ctx, errChan)
+	for {
+		select {
+		case err := <-errChan:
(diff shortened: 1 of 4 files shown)


bk201 (approved):

pkg/udev/uevent.go
```diff
@@ -19,28 +19,45 @@ import (
...
 }
 
 func (u *Udev) Monitor(ctx context.Context) {
-	u.startOnce.Do(func() {
-		u.monitor(ctx)
-	})
+	// we need to respawn the monitor with any error.
+	// because any error will break the monitor loop.
+	errChan := make(chan error)
+	go u.sapwnMonitor(ctx, errChan)
```
tserong, line 43:
```suggestion
	go u.spawnMonitor(ctx, errChan)
```

pkg/udev/uevent.go
```diff
@@ -19,28 +19,45 @@ import (
...
+
 }
 
-func (u *Udev) monitor(ctx context.Context) {
+func (u *Udev) sapwnMonitor(ctx context.Context, errChan chan error) {
```
tserong, line 47:
```suggestion
func (u *Udev) spawnMonitor(ctx context.Context, errChan chan error) {
```

tserong (approved):
I found two small typos, other than that it looks fine (i.e. logic seems sensible).

I was going to say that the thing I don't understand is _why_ it terminates in the first place if there's an error (our code has an infinite loop, right?) but then I went looking at https://github.com/harvester/node-disk-manager/blob/44f601cfff1fc9f7768504ae1a2bff85e119ec51/vendor/github.com/pilebones/go-udev/netlink/conn.go#L168-L175 and discovered the goroutine in `Monitor` will just stop if there's any kind of error.

Vicente-Cheng:
Hi @tserong, thanks for the review!

>I was going to say that the thing I don't understand is why it terminates in the first place if there's an error (our code has an infinite loop, right?) but then I went looking at ...

As you read the `go-udev` behavior, they will break loop with any error.
That means even if we do not stop the goroutine, we can not receive any udev event.

So that is why we need to respawn it (means creating a new monitor for incoming udev event)

pkg/udev/uevent.go
Vicente-Cheng: nice catch! thanks

Vicente-Cheng:
@mergifyio backport v0.6.x

Vicente-Cheng:
@mergifyio backport v0.5.x

pkg/udev/uevent.go
```diff
@@ -55,16 +72,24 @@ func (u *Udev) monitor(ctx context.Context) {
...
 	// Handling message from udev queue
 	for {
 		select {
 		case uevent := <-uqueue:
 			u.ActionHandler(uevent)
-		case err := <-errors:
-			logrus.Errorf("failed to parse udev event, error: %s", err.Error())
+		case err := <-errChan:
+			errors <- err
+			return
```
w13915984028, line 92: As a new `return` is added here, and L94 has a `close(quit)`,

It seems a `defer close(quit)` is needed in  L77, the L79 `if u.injectError` block will also leave `quit` un-closed.

Vicente-Cheng: fixed on https://github.com/harvester/node-disk-manager/pull/79

nice catch! thanks

PR #80, opened 2024-01-28
Add ns related funcs

Vicente-Cheng:
**Problem:**
Move ns related functions locally instead of extra packages

**Solution:**
Implement our own ns related ns functions

**Related Issue:**

**Test plan:**
Make sure everything will be fine after that.

### DIFF
--- a/pkg/utils/command.go
+++ b/pkg/utils/command.go
@@ -1,0 +1,83 @@
+package utils
+
+import (
+	"bytes"
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	NSBinary          = "nsenter"
+	cmdTimeoutDefault = 180 * time.Second // 3 minutes by default
+	cmdTimeoutNone    = 0 * time.Second   // no timeout
+)
+
+type Executor struct {
+	namespace  string
+	cmdTimeout time.Duration
+}
+
+func NewExecutor() *Executor {
+	return &Executor{
+		namespace:  "",
+		cmdTimeout: cmdTimeoutDefault,
+	}
+}
+
+func NewExecutorWithNS(ns string) (*Executor, error) {
+	exec := NewExecutor()
+	exec.namespace = ns
+
+	// test if nsenter is available
+	if _, err := execute(NSBinary, []string{"-V"}, cmdTimeoutNone); err != nil {
+		return nil, errors.Wrap(err, "cannot find nsenter for namespace switching")
+	}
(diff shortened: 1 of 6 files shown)


bk201:
Did we see any issue with importing `github.com/longhorn/go-iscsi-helper` before?

Vicente-Cheng:
Because of various improvements in Longhorn Common Library (after v1.6.0), some related functions have been removed.

We only need a few parts, so I prefer to maintain ourselves.
That will also give us more flexibility.

pkg/utils/command.go
```diff
@@ -0,0 +1,112 @@
...
+	var output, stderr bytes.Buffer
+	cmd.Stdout = &output
+	cmd.Stderr = &stderr
+
+	if err = cmd.Run(); err != nil {
+		return "", errors.Wrapf(err, "failed to execute: %v %v, output %s, stderr %s",
+			command, args, output.String(), stderr.String())
+	}
+	return output.String(), nil
+}
```
tserong, line 112: I kinda feel like `executeWithoutTimeout()` could somehow be merged with `executeWithTimeout()`, given there's a lot of similarity, I'm just not sure of the best way to do it :-)

Also, it looks like we don't currently use any of the timeout stuff anyway. Do we need to include it at all?

pkg/utils/command.go
```diff
@@ -0,0 +1,112 @@
...
+
+func NewExecutor() *Executor {
+	return &Executor{
+		namespace:  "",
+		cmdTimeout: cmdTimeoutDefault,
+	}
+}
+
+func (exec *Executor) PatchNS(ns string) error {
+	exec.namespace = ns
```
tserong, line 32: Could this be merged into `NewExecutor()`, so we'd call `executor := util.NewExecutor(ns)` instead of the two-line form (`executor := util.NewExecutor() ; executor.PatchNS(ns)`)?

pkg/utils/command.go
Vicente-Cheng: good suggestions, trying to do a new implement for it
idea from https://github.com/golang/go/issues/9580

pkg/utils/command.go
Vicente-Cheng: I want to give a way to create a non-ns executor.
Add another init func called `NewExecutorWithNS`. WDYT?

bk201 (approved):

tserong (approved):

FrankYang0529 (approved):
LGTM. Thanks.

PR #81, opened 2024-01-29
controller: should respect the StorageReserved setting of Longhorn

Vicente-Cheng:
**Problem:**
Should update the StorageReserved for extra disks.
before fixing (this is disk related info on `nodes.longhorn.io`)
```
spec:
  allowScheduling: true
  disks:
    a827440145f11f652f00f2a6ae656ed0:
      allowScheduling: true
      diskType: filesystem
      evictionRequested: false
      path: /var/lib/harvester/extra-disks/a827440145f11f652f00f2a6ae656ed0
      storageReserved: 0
      tags: []
```

After fixing
```
spec:
  allowScheduling: true
  disks:
    a827440145f11f652f00f2a6ae656ed0:
      allowScheduling: true
      diskType: filesystem
      evictionRequested: false
      path: /var/lib/harvester/extra-disks/a827440145f11f652f00f2a6ae656ed0
      storageReserved: 3221225472
      tags: []
```

**Solution:**
get the Longhorn storage reserved settings to calculate the correct value

**Related Issue:**
https://github.com/harvester/harvester/issues/5042

**Test plan:**
make sure the `storageReserved` is the correct value. (default is 30%)

### DIFF
--- a/deploy/charts/harvester-node-disk-manager/templates/rbac.yaml
+++ b/deploy/charts/harvester-node-disk-manager/templates/rbac.yaml
@@ -15,7 +15,7 @@ rules:
     resources: [ "blockdevices" ]
     verbs: [ "*" ]
   - apiGroups: [ "longhorn.io" ]
-    resources: [ "nodes" ]
+    resources: [ "nodes", "settings" ]
     verbs: [ "get", "list", "watch", "update", "patch" ]
   - apiGroups: [ "" ]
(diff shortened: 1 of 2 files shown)


Vicente-Cheng:
Thanks @votdev for reviewing it.

I thought maybe we did not need this PR.  Confirming with LH team.

Vicente-Cheng:
drop this because we do not need storage reserved as default for extra disks

PR #82, opened 2024-03-03
Add longhorn v2 provisioner

Vicente-Cheng:
**Problem:**
Add capability to provision Longhorn v2 engine backend disk

**Solution:**
Add capability to provision Longhorn v2 engine backend disk

**Related Issue:**

**Test plan:**
TBD

**Tasks:**
- [ ] add integration test

### DIFF
--- a/pkg/controller/blockdevice/controller.go
+++ b/pkg/controller/blockdevice/controller.go
@@ -810,16 +810,6 @@ func convertFSInfoToString(fsInfo *block.FileSystemInfo) string {
 	return fmt.Sprintf("mountpoint: %s, fsType: %s", fsInfo.MountPoint, fsInfo.Type)
 }
 
-func removeUnNeeded[T string | int](x []T, y []T) []T {
-	result := make([]T, 0)
-	for _, item := range x {
-		if !slices.Contains(y, item) {
-			result = append(result, item)
-		}
-	}
-	return result
-}
-
 func generateDiskSpec(provisioner string) longhornv1.DiskSpec {
 	diskType := longhornv1.DiskTypeFilesystem
 	if provisioner == "longhornv2" {
--- a/pkg/controller/blockdevice/controller.go
+++ b/pkg/controller/blockdevice/controller.go
@@ -30,6 +30,7 @@ import (
 
 const (
 	blockDeviceHandlerName = "harvester-block-device-handler"
+	defaultProvisioner     = "longhornv1"
 )
 
 // semaphore is a simple semaphore implementation in channel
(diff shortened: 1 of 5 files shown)


tserong (commented):
LGTM. I note this PR is marked "draft" - are there more changes to come here?

Vicente-Cheng:
Hi @tserong,

Thanks for the quick review. 
After discussion, we might not make the longhorn v2 engine in v1.3.0.

So this change will still draft here until v1.3.0 release.
I will add the integration test, and maybe some minor change once I got any issue with integration test (CI).

pkg/controller/blockdevice/controller.go
```diff
@@ -212,12 +213,16 @@ func (c *Controller) OnBlockDeviceChange(_ string, device *diskv1.BlockDevice) (
 	if devPath == "" {
 		return nil, fmt.Errorf("failed to resolve persistent dev path for block device %s", device.Name)
 	}
+	if device.Spec.Provisioner == "" {
```
innobead, line 216: nit: Do we have a mutator mechanism to add the default value? If yes, we can prevent this sort of non-biz code in controller.

pkg/controller/blockdevice/controller.go
```diff
@@ -461,13 +467,14 @@ func (c *Controller) provisionDeviceToNode(device *diskv1.BlockDevice) error {
...
 	nodeCpy := node.DeepCopy()
-	diskSpec := longhornv1.DiskSpec{
-		Path:              extraDiskMountPoint(device),
-		AllowScheduling:   true,
-		EvictionRequested: false,
-		StorageReserved:   0,
-		Tags:              device.Spec.Tags,
+
+	diskSpec := generateDiskSpec(device.Spec.Provisioner)
+	if device.Spec.Provisioner == defaultProvisioner {
```
innobead, line 472: nit: suggest handling each case explicitly to prevent using `else` for v2. Can use switch.

pkg/utils/utils.go
```diff
@@ -175,3 +176,14 @@ func CallerWithCondLock[T any](cond *sync.Cond, f func() T) T {
 	defer cond.L.Unlock()
 	return f()
 }
+
+// DoCommand executes a command and returns the stdout, stderr and error
+func DoCommand(cmdString string) (string, string, error) {
```
innobead, line 181: Nit: This function is quite general. Not sure if there is an existing lib able to use or even we can add this to common lib for other repos to reuse later if appropriate.  

Can extend it to pass io.Writer out and err instead of hard-coded bytes buffer here only. 

If for testing only, why not just add this to a related test package folder?

innobead (commented):
In general, LGTM. Just a few comments.

PR #100, opened 2024-05-09
scanner: add more logs

Vicente-Cheng:
**Problem:**
We encounter some weird issue that the ID is changed after reboot.

**Solution:**
 Add these logs would help us to verify easily.

**Related Issue:**

**Test plan:**
None

### DIFF
--- a/pkg/controller/blockdevice/scanner.go
+++ b/pkg/controller/blockdevice/scanner.go
@@ -98,6 +98,11 @@ func (s *Scanner) collectAllDevices() []*deviceWithAutoProvision {
 			logrus.Infof("Skip adding non-identifiable block device /dev/%s", disk.Name)
 			continue
 		}
+		logrus.Infof("Adding the disk with block device /dev/%s, id(Name): %s on node %s", disk.Name, bd.Name, s.NodeName)
+		logrus.Infof("  - wwn: %v", bd.Status.DeviceStatus.Details.WWN)
+		logrus.Infof("  - vendor: %v", bd.Status.DeviceStatus.Details.Vendor)
+		logrus.Infof("  - model: %v", bd.Status.DeviceStatus.Details.Model)
+		logrus.Infof("  - SerialNumber: %v", bd.Status.DeviceStatus.Details.SerialNumber)
 		autoProv := s.ApplyAutoProvisionFiltersForDisk(disk)
 		allDevices = append(allDevices, &deviceWithAutoProvision{bd: bd, AutoProvisioned: autoProv})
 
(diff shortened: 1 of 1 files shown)


pkg/controller/blockdevice/scanner.go
```diff
@@ -98,6 +98,12 @@ func (s *Scanner) collectAllDevices() []*deviceWithAutoProvision {
 			logrus.Infof("Skip adding non-identifiable block device /dev/%s", disk.Name)
 			continue
 		}
+		logrus.Infof("Adding the disk with block device /dev/%s, id(Name): %s on node %s", disk.Name, bd.Name, s.NodeName)
+		logrus.Infof("Some disk details like below:")
```
tserong, line 102: I'd suggest not including the above line (just because I think it's clear that the information below applies to the disk being added)

tserong (approved):
LGTM (just one tiny suggestion)

pkg/controller/blockdevice/scanner.go
Vicente-Cheng: good suggestion, updated!

bk201 (approved):
lgtm

pkg/controller/blockdevice/scanner.go
```diff
@@ -98,6 +98,11 @@ func (s *Scanner) collectAllDevices() []*deviceWithAutoProvision {
 			logrus.Infof("Skip adding non-identifiable block device /dev/%s", disk.Name)
 			continue
 		}
+		logrus.Infof("Adding the disk with block device /dev/%s, id(Name): %s on node %s", disk.Name, bd.Name, s.NodeName)
```
votdev, line 101: It's already too late now that the PR has been merged, but this question nevertheless. 

Does it not make sense to write this information in one line? The output in the log does not always have to look like in the code. In the meantime in which these output lines are written to the log, other messages can also be written. This can make the output quite confusing. Shouldn't `logrus.WithFields` be used here?

Vicente-Cheng: Yeah, I agree that.
We will use `WithFields` after introducing the log format.

I thought we had a similar issue with this (for the whole harvester components).

votdev: > I thought we had a similar issue with this (for the whole harvester components).

Yes, see [here](https://github.com/harvester/harvester/issues/5402).

Vicente-Cheng: Exactly this one. I would like to refine all the logs when processing this.

Vicente-Cheng:

PR #101, opened 2024-05-09
ci: move drone to github action

Vicente-Cheng:
**Problem:**
move image related work to github action

**Solution:**
migrate from drone to github

**Related Issue:**
https://github.com/harvester/harvester/issues/5327

**Test plan:**
None


scripts/build
```diff
@@ -13,8 +13,6 @@ fi
 LINKFLAGS="-X github.com/harvester/node-disk-manager/pkg/version.Version=$VERSION
            -X github.com/harvester/node-disk-manager/pkg/version.GitCommit=$COMMIT $LINKFLAGS"
 
-CGO_ENABLED=0 go build -ldflags "$LINKFLAGS $OTHER_LINKFLAGS" -o bin/node-disk-manager
-if [ "$CROSS" = "true" ] && [ "$ARCH" = "amd64" ]; then
-    GOOS=darwin go build -ldflags "$LINKFLAGS" -o bin/node-disk-manager-darwin
-    GOOS=windows go build -ldflags "$LINKFLAGS" -o bin/node-disk-manager-windows
-fi
```
brandboat, line 20: So we don't need this anymore ? I saw this in node-manager too

Vicente-Cheng: Yeah, I think so. The Node-disk-manager did not run alone on windows or macos.
Means we do not need these two.

tserong (approved):
LGTM!

(I wondered why this one didn't add a `.github/workflows/pull_request.yaml` like https://github.com/harvester/harvester-csi-driver/pull/38 has, but then found we've already got `.github/workflows/basic-ci.yaml` which runs on a self-hosed runner, so it looks like the PR case is already handled)

Vicente-Cheng:
> LGTM!
> 
> (I wondered why this one didn't add a `.github/workflows/pull_request.yaml` like [harvester/harvester-csi-driver#38](https://github.com/harvester/harvester-csi-driver/pull/38) has, but then found we've already got `.github/workflows/basic-ci.yaml` which runs on a self-hosed runner, so it looks like the PR case is already handled)

yes, we already have the PR checking CI. It does more than just build, so I did not add `.github/workflows/pull_request.yaml`

bk201 (approved):
LGTM

Vicente-Cheng:

PR #102, opened 2024-05-10
controller: improve the mechanism of disk removing

Vicente-Cheng:
**Problem:**
When the disk is gone, we cannot remove it on Harvester UI.
That's a regression after we introduce the inactive/corrupted concept.

**Solution:**
Improve the removing disk mechanism.

**Related Issue:**
https://github.com/harvester/harvester/issues/5726

**Test plan:**
1. Create 3 node Harvester cluster
2. add an extra disk and provision it as Longhorn disk
3. remove the above disk (physically). (If you use kvm, try virtctl command with attach-device/detach-device)
4. Remove this provision disk on the Harvester UI
5. make sure both Harvester/Longhorn UI removed the above disk

**Nice to Have:**
- [x] Add new integration test for the above case

### DIFF
--- a/pkg/controller/blockdevice/controller.go
+++ b/pkg/controller/blockdevice/controller.go
@@ -573,23 +573,36 @@ func (c *Controller) unprovisionDeviceFromNode(device *diskv1.BlockDevice) error
 		return nil
 	}
 
+	isUnprovisioning := false
+	for _, tag := range device.Status.Tags {
+		if tag == utils.DiskRemoveTag {
+			isUnprovisioning = true
+			break
+		}
+	}
+
 	// for inactive/corrupted disk, we could remove it from node directly
-	if isValidateToDelete(diskToRemove) &&
+	if isUnprovisioning && isValidateToDelete(diskToRemove) &&
 		(device.Status.State == diskv1.BlockDeviceInactive || device.Status.DeviceStatus.FileSystem.Corrupted) {
 		logrus.Infof("disk (%s) is inactive or corrupted, remove it from node directly", device.Name)
+		// handle mountpoint first
+		filesystem := c.BlockInfo.GetFileSystemInfoByDevPath(device.Status.DeviceStatus.DevPath)
+		if filesystem != nil && filesystem.MountPoint != "" {
+			timeout := 30 * time.Second
+			if err := utils.ForceUmountWithTimeout(filesystem.MountPoint, timeout); err != nil {
+				logrus.Warnf("Force umount %v error: %v", filesystem.MountPoint, err)
+			}
+			// reset related fields
+			c.updateDeviceFileSystem(device, device.Status.DeviceStatus.DevPath)
+			device.Spec.Tags = []string{}
+			device.Status.Tags = []string{}
+		}
+		// remove the disk from node
 		if err := removeDiskFromNode(); err != nil {
 			return err
 		}
 		updateProvisionPhaseUnprovisioned()
 		return nil
-	}
-
-	isUnprovisioning := false
-	for _, tag := range diskToRemove.Tags {
-		if tag == utils.DiskRemoveTag {
-			isUnprovisioning = true
-			break
-		}
 	}
 
 	if isUnprovisioning {
--- a/pkg/controller/blockdevice/controller.go
+++ b/pkg/controller/blockdevice/controller.go
@@ -188,14 +188,41 @@ func Register(
 	return nil
 }
 
+func (c *Controller) handleDeviceRemove(device *diskv1.BlockDevice) error {
+	logrus.Infof("Prepare to stop provisioning device %s to node %s", device.Name, c.NodeName)
+	if err := c.unprovisionDeviceFromNode(device); err != nil {
+		err := fmt.Errorf("failed to stop provisioning device %s to node %s: %w", device.Name, c.NodeName, err)
+		logrus.Warnf("Removing disk %v error: %v", device.Name, err)
+		return err
+	}
+	return nil
+}
+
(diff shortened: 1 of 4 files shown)


Vicente-Cheng:
> I think I understand this :-) and it looks reasonable. I just have a couple of questions:
> 
> 1. It looks like if the call to `utils.ForceUmountWithTimeout()` fails due to timeout, then the disk won't be removed from the node. Given that we _do_ want to remove  inactive or corrupted disks from the node, would it make sense to ignore timeout failures and remove the disk anyway?

Ah... good catch, I should not return. Maybe just log and continue the remove op with longhorn node.
Let me update again...

> 2. I gave this a rough test by attaching a disk to harvester running in a VM, then I let it be provisioned, detached the disk, rebooted that node, waited for it to come up showing a failed disk, and then finally I updated node-disk-manager to a build I did of this PR. The failed/missing disk was automatically removed from the node (when viewed via the Harvester UI), and from Longhorn. So that all works fine. But, it's still present as a blockdevice CR. Is that correct? Or should the CR be deleted as well?

Yeah, currently, the blockdevice CR needs to be deleted manually. There are some reasons I did not delete blockdevice CRD with any delete operations.

1. for hotplug case, the blockdevice will change the state from inactive -> active.
2. for corruption cases, the block device needs to keep recording the corruption flag. Then, we could try to repair and reuse manually.


But I also wonder we need to refine the whole flow. I will create a HEP to redefine the NDM on v1.4.0 to decompose the component like the provisioned. I thought we could discuss more at that time.

tserong (approved):

pkg/controller/blockdevice/controller.go
```diff
@@ -188,14 +188,41 @@ func Register(
...
+
 // OnBlockDeviceChange watch the block device CR on change and performing disk operations
 // like mounting the disks to a desired path via ext4
 func (c *Controller) OnBlockDeviceChange(_ string, device *diskv1.BlockDevice) (*diskv1.BlockDevice, error) {
-	if device == nil || device.DeletionTimestamp != nil || device.Spec.NodeName != c.NodeName || device.Status.State == diskv1.BlockDeviceInactive {
+	if device == nil || device.DeletionTimestamp != nil || device.Spec.NodeName != c.NodeName {
 		return nil, nil
 	}
 
+	// handle remove device no matter interactive or corrupted, we will set `device.Spec.FileSystem.Provisioned` to false
```
votdev, line 208:
```suggestion
	// handle remove device no matter inactive or corrupted, we will set `device.Spec.FileSystem.Provisioned` to false
```

pkg/controller/blockdevice/controller.go
```diff
@@ -188,14 +188,41 @@ func Register(
...
+	if !device.Spec.FileSystem.Provisioned && device.Status.ProvisionPhase != diskv1.ProvisionPhaseUnprovisioned {
+		deviceCpy := device.DeepCopy()
+		if err := c.handleDeviceRemove(deviceCpy); err != nil {
+			diskv1.DiskAddedToNode.SetError(deviceCpy, "", err)
+			diskv1.DiskAddedToNode.SetStatusBool(deviceCpy, false)
+			c.Blockdevices.EnqueueAfter(c.Namespace, device.Name, jitterEnqueueDelay())
+		}
+		if !reflect.DeepEqual(device, deviceCpy) {
+			logrus.Debugf("Update block device %s after removeing", device.Name)
```
votdev, line 217:
```suggestion
			logrus.Debugf("Update block device %s after removing", device.Name)
```

pkg/utils/utils.go
```diff
@@ -121,6 +124,20 @@ func UmountDisk(path string) error {
 	return os.NewSyscallError("umount", err)
 }
 
+func ForceUmountWithTimeout(path string, timeout time.Duration) error {
+	isHostProcMounted, err := IsHostProcMounted()
+	if err != nil {
+		return err
+	}
+	if isHostProcMounted {
+		_, err := executeOnHostNamespaceWithTimeout("umount", []string{path, "-f"}, timeout)
```
votdev, line 133: I think it is best practice to set the command options first and at last the device file.
```suggestion
		_, err := executeOnHostNamespaceWithTimeout("umount", []string{"-f", path}, timeout)
```

pkg/utils/utils.go
```diff
@@ -157,6 +174,16 @@ func executeOnHostNamespace(cmd string, args []string) (string, error) {
 	return executor.Execute(cmd, args)
 }
 
+func executeOnHostNamespaceWithTimeout(cmd string, args []string, cmdTimeout time.Duration) (string, error) {
```
votdev, line 177: I know it's no fun and it's a chore to document functions, but I'm wondering what is returned by the function. In case of an error I see an empty string, but on success? I can only answer this question if I look at what executor.Execute returns. As a developer I should be able to use the function `executeOnHostNamespaceWithTimeout` without having to do a code review and research the functions used.

votdev (changes requested):

pkg/controller/blockdevice/controller.go
Vicente-Cheng: updated, thanks!

pkg/controller/blockdevice/controller.go
Vicente-Cheng: ditto

pkg/utils/utils.go
Vicente-Cheng: Nice catch, updated!

votdev (approved):

Vicente-Cheng:
@mergifyio backport v0.6.x

PR #103, opened 2024-05-15
fix the duplicated wwn case

Vicente-Cheng:
**Problem:**
We have problems with duplicated WWN as https://github.com/harvester/harvester/issues/5446#issuecomment-2112177685

**Solution:**
Should skip the handling with the udev ADD event

**Related Issue:**
https://github.com/harvester/harvester/issues/4494

**Test plan:**
Please check the above issue.

### DIFF
--- a/pkg/controller/blockdevice/scanner.go
+++ b/pkg/controller/blockdevice/scanner.go
@@ -98,7 +98,7 @@ func (s *Scanner) collectAllDevices() []*deviceWithAutoProvision {
 			logrus.Infof("Skip adding non-identifiable block device /dev/%s", disk.Name)
 			continue
 		}
-		logrus.Infof("Adding the disk with block device /dev/%s, id(Name): %s on node %s", disk.Name, bd.Name, s.NodeName)
+		logrus.Infof("Detected the disk with block device /dev/%s, id(Name): %s on node %s", disk.Name, bd.Name, s.NodeName)
 		logrus.Infof("  - wwn: %v", bd.Status.DeviceStatus.Details.WWN)
 		logrus.Infof("  - vendor: %v", bd.Status.DeviceStatus.Details.Vendor)
 		logrus.Infof("  - model: %v", bd.Status.DeviceStatus.Details.Model)
@@ -125,8 +125,18 @@ func (s *Scanner) collectAllDevices() []*deviceWithAutoProvision {
 
 func (s *Scanner) handleExistingDev(oldBd *diskv1.BlockDevice, newBd *diskv1.BlockDevice, autoProvisioned bool) {
 	if isDevPathChanged(oldBd, newBd) {
-		logrus.Debugf("Enqueue block device %s for device path change", newBd.Name)
-		s.Blockdevices.Enqueue(s.Namespace, newBd.Name)
+		// the Dev Path changed only when the device is temporarily gone and back.
+		if oldBd.Status.State != diskv1.BlockDeviceInactive {
+			logrus.Warnf("Device Path should not change with the active device %s, new device path: %s", newBd.Name, newBd.Status.DeviceStatus.DevPath)
+		} else {
+			logrus.Infof("The inactive block device %s with wwn %s is coming back", newBd.Name, newBd.Status.DeviceStatus.Details.WWN)
+			oldBd.Status.State = diskv1.BlockDeviceActive
+			oldBd.Status.DeviceStatus.DevPath = newBd.Status.DeviceStatus.DevPath
+			if _, err := s.Blockdevices.Update(oldBd); err != nil {
+				logrus.Errorf("Update device %s status error", oldBd.Name)
+				s.Blockdevices.Enqueue(s.Namespace, newBd.Name)
+			}
+		}
 	} else if isDevAlreadyProvisioned(newBd) {
 		logrus.Debugf("Skip the provisioned device: %s", newBd.Name)
 	} else if s.NeedsAutoProvision(oldBd, autoProvisioned) {
@@ -196,7 +206,6 @@ func (s *Scanner) scanBlockDevicesOnNode() error {
 				continue
 			}
 			logrus.Infof("Create new device %s with wwn: %s", bd.Name, bd.Status.DeviceStatus.Details.WWN)
-			// persist newly detected block device
 			if _, err := s.SaveBlockDevice(bd, autoProvisioned); err != nil && !errors.IsAlreadyExists(err) {
 				return err
 			}
@@ -267,7 +276,7 @@ func (s *Scanner) ApplyAutoProvisionFiltersForDisk(disk *block.Disk) bool {
 
 // SaveBlockDevice persists the blockedevice information.
 func (s *Scanner) SaveBlockDevice(bd *diskv1.BlockDevice, autoProvisioned bool) (*diskv1.BlockDevice, error) {
-	curBd, err := s.Blockdevices.Get(bd.Namespace, bd.Name, metav1.GetOptions{})
+	_, err := s.Blockdevices.Get(bd.Namespace, bd.Name, metav1.GetOptions{})
 	if err != nil {
 		if errors.IsNotFound(err) {
 			if autoProvisioned {
@@ -279,9 +288,8 @@ func (s *Scanner) SaveBlockDevice(bd *diskv1.BlockDevice, autoProvisioned bool) 
 		}
 		return nil, err
 	}
-	logrus.Infof("The inactive block device %s with wwn %s is coming back", bd.Name, bd.Status.DeviceStatus.Details.WWN)
-	curBd.Status.State = diskv1.BlockDeviceActive
-	return s.Blockdevices.Update(curBd)
+	logrus.Warnf("Should be handled by existing device, coming bd: %v", bd)
+	return nil, nil
 }
 
 // NeedsAutoProvision returns true if the current block device needs to be auto-provisioned.
@@ -296,9 +304,9 @@ func (s *Scanner) NeedsAutoProvision(oldBd *diskv1.BlockDevice, autoProvisionPat
 
 // isDevPathChanged returns true if the device path has changed.
 //
-// When reboot, device path might change but controller cannot detect
-// it during the reconciliation. We explicitly check and update the value
-// when scanner startup.
+// The device path changed on when the device is temporarily gone and back.
+// We init the device status when the first time the device is found.
+// If the device is active, the device path should not change.
 func isDevPathChanged(oldBd *diskv1.BlockDevice, newBd *diskv1.BlockDevice) bool {
 	return oldBd.Status.DeviceStatus.DevPath != newBd.Status.DeviceStatus.DevPath
 }
--- a/pkg/udev/uevent.go
+++ b/pkg/udev/uevent.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/pilebones/go-udev/netlink"
 	"github.com/sirupsen/logrus"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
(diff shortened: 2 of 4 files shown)


pkg/controller/blockdevice/scanner.go
```diff
@@ -125,8 +125,18 @@ func (s *Scanner) collectAllDevices() []*deviceWithAutoProvision {
 
 func (s *Scanner) handleExistingDev(oldBd *diskv1.BlockDevice, newBd *diskv1.BlockDevice, autoProvisioned bool) {
 	if isDevPathChanged(oldBd, newBd) {
-		logrus.Debugf("Enqueue block device %s for device path change", newBd.Name)
-		s.Blockdevices.Enqueue(s.Namespace, newBd.Name)
+		// the Dev Path changed only when the device is temporarily gone and back.
```
tserong, line 128: There's a comment on the `isDevPathChanged()` function which says "When reboot, device path might change", but the comment in the line above here says the De Path  only changes when the device goes away temporarily, which seems to conflict with the other comment. Which of these is correct?

pkg/udev/uevent.go
```diff
@@ -155,11 +153,8 @@ func (u *Udev) ActionHandler(uevent netlink.UEvent) {
 }
 
 // AddBlockDevice add new block device and partitions by watching the udev add action
-func (u *Udev) AddBlockDevice(device *v1beta1.BlockDevice, autoProvisioned bool) {
-	_, err := u.scanner.SaveBlockDevice(device, autoProvisioned)
-	if err != nil && !apierrors.IsAlreadyExists(err) {
-		logrus.Errorf("failed to save block device %s, error: %s", device.Name, err.Error())
-	}
+func (u *Udev) AddBlockDevice(_ *v1beta1.BlockDevice) {
+	// the behavior will execute when receive the ADD action from udev
```
tserong, line 157: This function appears to be not called anymore. Can we delete it entirely?

tserong (commented):
I actually don't yet understand how this change fixed the duplicate wwn problem, so I'll have to come back to it on Monday :-) but in the meantime I have a couple of questions.

pkg/udev/uevent.go
votdev: `AddBlockDevice` is not called and it's empty.

pkg/controller/blockdevice/scanner.go
```diff
@@ -125,8 +125,18 @@ func (s *Scanner) collectAllDevices() []*deviceWithAutoProvision {
...
+		if oldBd.Status.State != diskv1.BlockDeviceInactive {
+			logrus.Warnf("Dev Path should not change for provisioned device %s, new dev path: %s", newBd.Name, newBd.Status.DeviceStatus.DevPath)
```
votdev, line 130: After removing the sda disk, i got this message:
```
Dev Path should not change for provisioned device 379bdfe9134b148a28fea81aabe91d09, new dev path: /dev/sdb
```
IMO this is misleading and is causing some confusion. I think the message should be more clear what is happening here and that the action is ignored.

pkg/udev/uevent.go
Vicente-Cheng: Yeah, now no one calls this function. The older caller is on https://github.com/harvester/node-disk-manager/blob/master/pkg/udev/uevent.go#L140

I keep this here because we could easily add any operations when we want to.
If you thought we could remove it directly, I could update.

pkg/controller/blockdevice/scanner.go
Vicente-Cheng: Yeah, I did not notice the original comment of `isDevPathChanged()`.
I thought the reboot cases did not cause different `DeviceStatus.DevPath` with the same device.

The `DeviceStatus.DevPath` is filled with first-time NDM brought up.
As we know, the device path should not change once it registers on the system.
The only chance is this device is gone and back.

So I will correct the `isDevPathChanged()` comment.
Thanks for it.

pkg/controller/blockdevice/scanner.go
Vicente-Cheng: I agree that will confuse the user.
How about?
```
Device Path should not change with the active device %s, new device path: %s", newBd.Name, newBd.Status.DeviceStatus.DevPath
```

pkg/udev/uevent.go
votdev: I usually want to keep existing code, even if it is not used. I am therefore always reminded to remove this code. The code could be found again later in the history and reactivated; but what developer thinks about looking in the history when searching for usable code?
But since the function is really empty and nobody has anything of it, I think it can be removed.

votdev:
Could you please change `Adding the disk with block device /dev/%s ...` at https://github.com/harvester/node-disk-manager/blob/969ccdc32fa1ba1a9bda27a9a2024f36dfcd300a/pkg/controller/blockdevice/scanner.go#L101
to `Found the disk with block device /dev/%s` or `Detected the disk with block device /dev/%s` because this code does only add the disk to a list that is processed later, but the log message might indicate erroneously that the disk is added to a database or something else.

The message `Create new device %s with wwn: ...` at https://github.com/harvester/node-disk-manager/blob/969ccdc32fa1ba1a9bda27a9a2024f36dfcd300a/pkg/controller/blockdevice/scanner.go#L198 should make it clear what has happened to the found/detected device.

votdev (approved):
Except my other comments it looks good.

```
time="2024-05-27T08:45:08Z" level=info msg="Adding the disk with block device /dev/sda, id(Name): 5b8ab6108a3d8ba74a252d1f6737d8bb on node harvester-node-0"
time="2024-05-27T08:45:08Z" level=info msg="  - wwn: 0x5000c50017654321"
time="2024-05-27T08:45:08Z" level=info msg="  - vendor: aaa"
time="2024-05-27T08:45:08Z" level=info msg="  - model: QEMU_HARDDISK"
time="2024-05-27T08:45:08Z" level=info msg="  - SerialNumber: 5000c50017654321"
time="2024-05-27T08:45:08Z" level=info msg="Create new device 5b8ab6108a3d8ba74a252d1f6737d8bb with wwn: 0x5000c50017654321"
time="2024-05-27T08:45:08Z" level=info msg="Add new block device 5b8ab6108a3d8ba74a252d1f6737d8bb with device: /dev/sda"
time="2024-05-27T08:45:08Z" level=info msg="Waiting new event to trigger..."
time="2024-05-27T08:47:06Z" level=info msg="Wake up scanner with add operation with blockdevice: 5670b08adefeff4182eac8c1d7e21178"
time="2024-05-27T08:47:06Z" level=info msg="scanner waked up, do scan..."
time="2024-05-27T08:47:06Z" level=info msg="Adding the disk with block device /dev/sda, id(Name): 5b8ab6108a3d8ba74a252d1f6737d8bb on node harvester-node-0"
time="2024-05-27T08:47:06Z" level=info msg="  - wwn: 0x5000c50017654321"
time="2024-05-27T08:47:06Z" level=info msg="  - vendor: aaa"
time="2024-05-27T08:47:06Z" level=info msg="  - model: QEMU_HARDDISK"
time="2024-05-27T08:47:06Z" level=info msg="  - SerialNumber: 5000c50017654321"
time="2024-05-27T08:47:06Z" level=info msg="Adding the disk with block device /dev/sdb, id(Name): 5670b08adefeff4182eac8c1d7e21178 on node harvester-node-0"
time="2024-05-27T08:47:06Z" level=info msg="  - wwn: 0x5000c50017654321"
time="2024-05-27T08:47:06Z" level=info msg="  - vendor: bbb"
time="2024-05-27T08:47:06Z" level=info msg="  - model: QEMU_HARDDISK"
time="2024-05-27T08:47:06Z" level=info msg="  - SerialNumber: 5000c50017654321"
time="2024-05-27T08:47:06Z" level=warning msg="Skip adding duplicated WWN device 0x5000c50017654321, device path: /dev/sdb"
```

pkg/controller/blockdevice/scanner.go
Vicente-Cheng: Updated!

Vicente-Cheng:
> Could you please change `Adding the disk with block device /dev/%s ...` at
> 
> https://github.com/harvester/node-disk-manager/blob/969ccdc32fa1ba1a9bda27a9a2024f36dfcd300a/pkg/controller/blockdevice/scanner.go#L101
> 
> 
> to `Found the disk with block device /dev/%s` or `Detected the disk with block device /dev/%s` because this code does only add the disk to a list that is processed later, but the log message might indicate erroneously that the disk is added to a database or something else.
> The message `Create new device %s with wwn: ...` at
> 
> https://github.com/harvester/node-disk-manager/blob/969ccdc32fa1ba1a9bda27a9a2024f36dfcd300a/pkg/controller/blockdevice/scanner.go#L198
> 
> should make it clear what has happened to the found/detected device.

Nice suggestion, updated.

pkg/udev/uevent.go
Vicente-Cheng: Sounds reasonable. I removed this function. Thanks!

pkg/controller/blockdevice/scanner.go
Vicente-Cheng: Update the comment!

tserong (approved):

bk201:
I reproduce a strange issue. It's probably not 100% related to this change, but it's worth checking.
Steps:
- Hot-plug 1 disk to node1 as sda. A BD is created.
- Add `sda`'s BD to longhorn.
- Hot-plug 1 new disk `sdb` with duplicated WWN to node1. No new BD is created because the disk's wwn is duplicated.
- Hot-plug 3 more disks (might not be necessary). I didn't add them to Longhorn.
- Reboot the node.
- After reboot, the original `sda` is still`sda`, but `sdb` is re-enumerated to `sde`.
- I can see BD for `sde` is not created. However, `sde` seems to be mounted as a longhorn disk.

[harvester-node-disk-manager-kvp6f.log](https://github.com/user-attachments/files/16076358/harvester-node-disk-manager-kvp6f.log)
[supportbundle_8b9e0c97-83f2-4b94-8922-b16a3481d1ce_2024-07-03T02-37-02Z.zip](https://github.com/user-attachments/files/16076362/supportbundle_8b9e0c97-83f2-4b94-8922-b16a3481d1ce_2024-07-03T02-37-02Z.zip)

Vicente-Cheng:
> I reproduce a strange issue. It's probably not 100% related to this change, but it's worth checking. Steps:
> 
> * Hot-plug 1 disk to node1 as sda. A BD is created.
> * Add `sda`'s BD to longhorn.
> * Hot-plug 1 new disk `sdb` with duplicated WWN to node1. No new BD is created because the disk's wwn is duplicated.
> * Hot-plug 3 more disks (might not be necessary). I didn't add them to Longhorn.
> * Reboot the node.
> * After reboot, the original `sda` is still`sda`, but `sdb` is re-enumerated to `sde`.
> * I can see BD for `sde` is not created. However, `sde` seems to be mounted as a longhorn disk.
> 
> [harvester-node-disk-manager-kvp6f.log](https://github.com/user-attachments/files/16076358/harvester-node-disk-manager-kvp6f.log) [supportbundle_8b9e0c97-83f2-4b94-8922-b16a3481d1ce_2024-07-03T02-37-02Z.zip](https://github.com/user-attachments/files/16076362/supportbundle_8b9e0c97-83f2-4b94-8922-b16a3481d1ce_2024-07-03T02-37-02Z.zip)

Yeah, That's a known issue.
This PR has handled the case that only the wwn is the same.

If we have two disks that are totally the same in wwn/model/vendor/serial number, currently, we would encounter this issue.
That's because they would use the same blockdevice CR, we could not identify them.

So the `/dev/sdb` in your case will use the blockdevice CR from `/dev/sda`
That why we see this result.

bk201:
Thanks for the reply. We can ignore this very corner case.

bk201 (approved):

Vicente-Cheng:

PR #109, opened 2024-06-05
Decompose the provisioner

Vicente-Cheng:
**Problem:**
To support various storage backends.

**Solution:**
Decompose the provisioner, we could define various behaviors for each provisioner

**Related Issue:**
TBD

**Test plan:**
Make sure all operations work as usual

**TASKS**
- [x] decompose the provisioner
- [x] create Longhornv1


Vicente-Cheng:
CodeFactor complains about files generated by wangler. (the redundant `return` of deepcopy func)

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_blockdevices.yaml
```diff
@@ -84,8 +84,24 @@ spec:
               nodeName:
                 description: name of the node to which the block device is attached
                 type: string
+              provisioner:
+                description: a string with the provisioner name, e.g. "Longhornv1,
+                  Longhornv2 or LocalLVM"
```
votdev, line 89: Is `LocalLVM` a typo? I am asking because some lines below in the properties you are using `Longhornv2 or LVM`.

manifests/crds/harvesterhci.io_blockdevices.yaml
```diff
@@ -84,6 +84,21 @@ spec:
               nodeName:
                 description: name of the node to which the block device is attached
                 type: string
+              provisioner:
+                description: a string with the provisioner name, e.g. "Longhornv1,
+                  Longhornv2 or LocalLVM"
+                properties:
+                  type:
+                    description: a string with the provisioner name, e.g. "Longhornv1,
+                      Longhornv2 or LVM"
```
votdev, line 93: `LocalLVM`or `LVM`?

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_blockdevices.yaml
```diff
@@ -274,6 +290,13 @@ spec:
                 - Unprovisioned
                 - Unprovisioning
                 type: string
+              provisioner:
+                description: a string with the provisioner name, e.g. "Longhornv1,
+                  Longhornv2 or LocalLVM"
```
votdev, line 295: LocalLVMor LVM?

manifests/crds/harvesterhci.io_blockdevices.yaml
```diff
@@ -275,6 +290,13 @@ spec:
                 - Unprovisioned
                 - Unprovisioning
                 type: string
+              provisioner:
+                description: a string with the provisioner name, e.g. "Longhornv1,
+                  Longhornv2 or LocalLVM"
```
votdev, line 295: LocalLVMor LVM?

pkg/apis/harvesterhci.io/v1beta1/types.go
```diff
@@ -62,6 +65,21 @@ type BlockDeviceStatus struct {
 
 	// The current Tags of the blockdevice
 	Tags []string `json:"tags,omitempty"`
+
+	// a string with the provisioner name, e.g. "Longhornv1, Longhornv2 or LocalLVM"
```
votdev, line 69: `LocalLVM` or `LVM`? Some lines below you are using `LVM`.

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_blockdevices.yaml
Vicente-Cheng: yeah, this is a typo. Sorry, I missed these comments. updated to all

pkg/provisioner/lvm.go
```diff
@@ -0,0 +1,309 @@
...
+type LVMProvisioner struct {
+	name      string
+	vgName    string
+	blockInfo block.Info
+	device    *diskv1.BlockDevice
+
+	semaphoreObj *Semaphore
+}
+
+func NewLVMProvisioner(vgName string, device *diskv1.BlockDevice, blockInfo block.Info, semaphoreObj *Semaphore) *LVMProvisioner {
```
Yu-Jack, line 23: Since these provisioners only have private variables (LVM & Longhorn), and you also have the interface here. You could do following changes:

1. Convert `LVMProvisioner` `LonghornV1Provisioner` to private.
2. Use interface as return in `NewLVMProvisioner` `NewLHV1Provisioner`

For example:

```suggestion
func NewLVMProvisioner(vgName string, device *diskv1.BlockDevice, blockInfo block.Info, semaphoreObj *Semaphore) Provisioner {
```

pkg/provisioner/lvm.go
```diff
@@ -0,0 +1,309 @@
...
+)
+
+type LVMProvisioner struct {
+	name      string
+	vgName    string
+	blockInfo block.Info
+	device    *diskv1.BlockDevice
+
+	semaphoreObj *Semaphore
+}
```
Yu-Jack, line 21: Just share another concept. We could use embedded struct to de-duplicate some codes.

```go
type provisioner struct {
	name      string
	blockInfo block.Info
	device    *diskv1.BlockDevice
}

func (p *provisioner) GetProvisionerName() string {
	return p.name
}

type LonghornV1Provisioner struct {
	*provisioner

	// other ... blablabla
}

type LVMProvisioner struct {
	*provisioner

	// other ... blablabla
}
```

Then you don't need to create common functions for each provisioner, such as `GetProvisionerName`. The reason I bring this is we'll have longhorn v2 engine in the future like you mentioned. It could reduce some repeated code if we need.

pkg/provisioner/common.go
```diff
@@ -0,0 +1,240 @@
...
+func convertMountStr(mountOP NeedMountUpdateOP) string {
+	switch mountOP {
+	case NeedMountUpdateNoOp:
+		return "No-Op"
+	case NeedMountUpdateMount:
+		return "Mount"
+	case NeedMountUpdateUnmount:
+		return "Unmount"
+	}
+	return "Unknown OP"
```
Yu-Jack, line 192: I think we could make use of default?

```suggestion
	case NeedMountUpdateUnmount:
		return "Unmount"
	default:
		return "Unknown OP"
	}
```

pkg/provisioner/lvm.go
```diff
@@ -0,0 +1,309 @@
...
+
+	args := []string{devPath}
+	_, err = executor.Execute("pvremove", args)
+	if err != nil {
+		return fmt.Errorf("pvremove failed. %v", err)
+	}
+	return nil
+}
+
+func doVGActivate() error {
```
Yu-Jack, line 296: I found these codes look similar, I think we could extract one more common function?

```go
func executeCommandWithNamespace(cmd string, args []string) error {
	ns := utils.GetHostNamespacePath(utils.HostProcPath)
	executor, err := utils.NewExecutorWithNS(ns)
	if err != nil {
		return fmt.Errorf("generate executor failed. %v", err)
	}

	_, err = executor.Execute(cmd, args)
	if err != nil {
		return fmt.Errorf("%s failed. %v", cmd, err)
	}
	return nil
}

func doPVCreate(devPath string) error {
	return executeCommandWithNamespace("pvcreate", []string{devPath})
}

func doVGCreate(devPath, vgName string) error {
	return executeCommandWithNamespace("vgcreate", []string{vgName, devPath})
}

func doVGActivate() error {
	return executeCommandWithNamespace("vgchange", []string{"--activate", "y"})
}
```

pkg/provisioner/lvm.go
Vicente-Cheng: good suggestion, updated!

pkg/provisioner/lvm.go
Vicente-Cheng: Thanks for the suggestion. That could help to reduce the duplicated common function! Updated!

pkg/provisioner/common.go
Vicente-Cheng: sure, updated!

pkg/provisioner/lvm.go
Vicente-Cheng: Thanks for the suggestion, updated!

tserong (commented):
Why is the provisioner info duplicated in both the CR spec and status sections? Doesn't this only need to be set in the spec?

pkg/controller/blockdevice/controller.go
```diff
@@ -188,455 +94,146 @@ func Register(
...
-		return err
-	}
-	diskv1.DeviceFormatting.SetError(device, "", nil)
-	diskv1.DeviceFormatting.SetStatusBool(device, false)
-	diskv1.DeviceFormatting.Message(device, "Done device ext4 filesystem formatting")
-	device.Status.DeviceStatus.FileSystem.LastFormattedAt = &metav1.Time{Time: time.Now()}
-	device.Status.DeviceStatus.Partitioned = false
-	device.Status.DeviceStatus.FileSystem.Corrupted = false
-	return nil
+	return nil, fmt.Errorf("unsupported provisioner type %s", provisionerType)
```
Yu-Jack, line 220: NIT: you could use switch default

Vicente-Cheng: updated, thanks!

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_blockdevices.yaml
```diff
@@ -23,7 +23,7 @@ spec:
     - jsonPath: .status.deviceStatus.details.deviceType
       name: Type
       type: string
-    - jsonPath: .spec.devPath
+    - jsonPath: .status.deviceStatus.devPath
```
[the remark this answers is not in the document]
Vicente-Cheng: The `spec.DevPath` means the dev path when we create the blockdevice CRD (means the first time and fixed)
The `status.deviceStatus.devPath` means the current dev path.

So I thought showing the `status.deviceStatus.devPath` seems more useful for us.

Yu-Jack (approved):
LGTM, thanks.

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_blockdevices.yaml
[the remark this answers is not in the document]
tserong: Yeah showing the actual current path makes sense.

I've just realised this (kinda) explains https://github.com/harvester/harvester/issues/6224 where there were discrepancies between `spec.DevPath` and `status.deviceStatus.devPath`. I mean, there never should have been three BDs that all had the same `status.deviceStatus.devPath`, and I assume that was (somehow) caused by multiple dodgy devices with the same WWN, but the fact that one is the path when the device was added, and one is the current path, at least sheds some light on the discrepancies I noticed in that issue.

Thinking aloud: Maybe we should just drop `spec.DevPath`? Does it help at all to know what the path was originally? Note: I'm _not_ suggesting we make any further changes to this PR in this area. Just an item for thought/consideration later.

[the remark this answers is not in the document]
Vicente-Cheng: >Maybe we should just drop spec.DevPath

+1, it's a redundant field for me with the current design.
We can do more cleanup for these fields. I plan to introduce a new naming mechanism to replace the current one in the next release (means v1.5). That we would be good at some devices w/o wwn. Maybe we could redesign these fields at that time.

We still need to check the WWN duplicated. But we could have better handling with those devices w/o wwn.

pkg/apis/harvesterhci.io/v1beta1/types.go
```diff
@@ -62,6 +64,25 @@ type BlockDeviceStatus struct {
...
+	// a string with the provisioner name, e.g. "Longhornv1, Longhornv2 or LVM"
+	Provisioner *ProvisionerStatus `json:"provisioner,omitempty"`
+}
+
+type ProvisionerInfo struct {
+	// a string with the provisioner name, e.g. "Longhornv1, Longhornv2 or LVM"
+	Type string `json:"type"`
+
+	// vgName is the volume group name for the provisioner (only worked for LVM)
+	VgName string `json:"vgName,omitempty"`
```
bk201, line 77: VgName is the required field for the LVM provisioner, but we may need more fields in the future. If every provisioner adds their fields in this structure, we'll have a lot of fields in the same level, and hard to tell which provisioner owns the field.

How about adding the provisioners' structures like this:

```
type ProvisionerInfo struct {
  LVMProvisionerInfo *LVMProvisionerInfo
  
  Provisioner2Info *Provisioner2Info

  Provisioner3Info *Provisioner3Info
}
```

Similar to the design of Volume source in the kubevirt VM: https://github.com/harvester/harvester/blob/88e05afc3a2a497b8b5cf76554d698a403ba7de1/vendor/kubevirt.io/api/core/v1/schema.go#L776

pkg/apis/harvesterhci.io/v1beta1/types.go
```diff
@@ -62,6 +64,25 @@ type BlockDeviceStatus struct {
 
 	// The current Tags of the blockdevice
 	Tags []string `json:"tags,omitempty"`
+
+	// a string with the provisioner name, e.g. "Longhornv1, Longhornv2 or LVM"
+	Provisioner *ProvisionerStatus `json:"provisioner,omitempty"`
```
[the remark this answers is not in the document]
Vicente-Cheng: agree, I want to use status to do some validation.
But it seems we will need a webhook. I could do the validation on the webhook.

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_blockdevices.yaml
```diff
@@ -274,6 +288,21 @@ spec:
...
+                description: a string with the provisioner name, e.g. "Longhornv1,
+                  Longhornv2 or LVM"
+                properties:
+                  type:
+                    description: a string with the provisioner name, e.g. "Longhornv1,
+                      Longhornv2 or LVM"
+                    type: string
+                  vgName:
+                    description: vgName is the volume group name for the provisioner
+                      (only worked for LVM)
```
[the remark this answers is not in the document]
Vicente-Cheng: no need to fix with the new provisioner field

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_blockdevices.yaml
```diff
@@ -84,8 +84,22 @@ spec:
...
                 type: string
+              provisioner:
+                properties:
+                  type:
+                    description: a string with the provisioner name, e.g. "Longhornv1,
+                      Longhornv2 or LVM"
+                    type: string
+                  vgName:
+                    description: vgName is the volume group name for the provisioner
+                      (only worked for LVM)
```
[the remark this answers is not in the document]
Vicente-Cheng: ditto

pkg/provisioner/common.go
```diff
@@ -0,0 +1,248 @@
...
+
+	diskv1 "github.com/harvester/node-disk-manager/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/node-disk-manager/pkg/block"
+)
+
+type NeedMountUpdateOP int8
+
+const (
+	TypeLonghornV1 = "Longhornv1"
+	TypeLonghornV2 = "Longhornv2"
```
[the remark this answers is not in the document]
Vicente-Cheng: thanks for the suggestion, updated!

pkg/apis/harvesterhci.io/v1beta1/types.go
Vicente-Cheng: Thanks for the suggestion, updated!

tserong (approved):

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_blockdevices.yaml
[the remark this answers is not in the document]
votdev: > Can these two values ever be different?

Yes. If you have a device file like `/dev/disk/by-path/pci-0000:00:10.0-scsi-0:0:0:0` this can change when the PCI slots are reordered, e.g. by adding new hardware.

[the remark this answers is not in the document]
Vicente-Cheng: Thanks @votdev,
Yeah, if we would like to keep the spec.devPath field. We should use a meaningful path like the above instead of the `/dev/sdx`. As I mentioned above, we could redesign the field and naming mechanism on v1.5.

pkg/apis/harvesterhci.io/v1beta1/types.go
```diff
@@ -64,6 +66,33 @@ type BlockDeviceStatus struct {
...
+	// +kubebuilder:validation:Required
+	VgName string `json:"vgName"`
+
+	// a string slice for the parameters
+	// +kubebuilder:validation:Optional
+	Parameters []string `json:"parameters,omitempty"`
+}
+
+type LonghornProvisionerInfo struct {
+	// a string with the volume name for the provisioner
```
votdev, line 91: Is this correct? Does `EngineVersion` not contain a engine version string?

pkg/provisioner/common.go
```diff
@@ -0,0 +1,248 @@
...
+)
+
+func (f NeedMountUpdateOP) Has(flag NeedMountUpdateOP) bool {
+	return f&flag != 0
+}
+
+type Provisioner interface {
+	// Format is the Prerequisites for the provisioner
+	// You should do the format-related operations (including mkfs, mount ...etc) here
+	// Return values: bool1: isFormatComplete, bool2: isNeedRequeue, error: error
```
votdev, line 40: Does this sound better? If changed, adapt the following occurrences as well.
```suggestion
	// Return values: bool1: isFormatComplete, bool2: isRequeueNeeded, error: error
```

pkg/provisioner/common.go
```diff
@@ -0,0 +1,248 @@
...
+	switch device.Status.DeviceStatus.Details.DeviceType {
+	case diskv1.DeviceTypeDisk:
+		// Disk naming priority.
+		// #1 WWN
+		// #2 filesystem UUID (UUID)
+		// #3 partition table UUID (PTUUID)
+		// #4 PtUUID as UUID to query disk info
+		//    (NDM might reuse PtUUID as UUID to format a disk)
+		if wwn := device.Status.DeviceStatus.Details.WWN; valueExists(wwn) {
+			if device.Status.DeviceStatus.Details.StorageController == string(diskv1.StorageControllerNVMe) {
```
votdev, line 213: Maybe add some links as references here:
- https://en.wikipedia.org/wiki/World_Wide_Name#Formats
- https://wiki.archlinux.org/title/Persistent_block_device_naming#by-id_and_by-path

pkg/provisioner/common.go
```diff
@@ -0,0 +1,248 @@
...
+				return filepath.EvalSymlinks("/dev/disk/by-id/nvme-" + wwn)
+			}
+			return filepath.EvalSymlinks("/dev/disk/by-id/wwn-" + wwn)
+		}
+		if fsUUID := device.Status.DeviceStatus.Details.UUID; valueExists(fsUUID) {
+			path, err := filepath.EvalSymlinks("/dev/disk/by-uuid/" + fsUUID)
```
votdev, line 219: Maybe add a link to a reference that explains how the device file should look like, e.g. https://wiki.archlinux.org/title/Persistent_block_device_naming#by-uuid

pkg/provisioner/common.go
```diff
@@ -0,0 +1,248 @@
...
+
+		if ptUUID := device.Status.DeviceStatus.Details.PtUUID; valueExists(ptUUID) {
+			path, err := block.GetDevPathByPTUUID(ptUUID)
+			if err != nil {
+				return "", err
+			}
+			if path != "" {
+				return path, nil
+			}
+			return filepath.EvalSymlinks("/dev/disk/by-uuid/" + ptUUID)
```
votdev, line 236: Add a reference link, e.g. https://wiki.archlinux.org/title/Persistent_block_device_naming#by-partuuid

pkg/provisioner/longhornv1.go
```diff
@@ -0,0 +1,457 @@
...
+		Tags:              tags,
+	}
+
+	// checked the case that longhorn node updated but blockdevice CRD is not updated
+	synced := false
+	provisioned := false
+	if disk, found := p.nodeObj.Spec.Disks[p.device.Name]; found {
+		synced = reflect.DeepEqual(disk, diskSpec)
+		if !synced {
+			logrus.Warnf("The disk spec should not different between longhorn node and blockdevice CRD, disk: %+v, diskSpec: %+v", disk, diskSpec)
```
votdev, line 87:
```suggestion
			logrus.Warnf("The disk spec should not differ between longhorn node and blockdevice CRD, disk: %+v, diskSpec: %+v", disk, diskSpec)
```

pkg/provisioner/longhornv1.go
```diff
@@ -0,0 +1,457 @@
...
+	p.cacheDiskTags.UpdateDiskTags(p.device.Name, p.device.Spec.Tags)
+	return false, nil
+}
+
+func (p *LonghornV1Provisioner) UnProvision() (bool, error) {
+	logrus.Infof("%s unprovisioning Longhorn block device %s", p.name, p.device.Name)
+
+	// inner functions
+	updateProvisionPhaseUnprovisioned := func() {
+		msg := fmt.Sprintf("Disk not in longhorn node `%s`", p.nodeObj.Name)
```
votdev, line 114: `Longhorn` is a product name, thus it should be title-case. Every occurrence across all files should be adapted.
```suggestion
		msg := fmt.Sprintf("Disk not in Longhorn node `%s`", p.nodeObj.Name)
```

pkg/provisioner/longhornv1.go
```diff
@@ -0,0 +1,457 @@
...
+
+func (p *LonghornV1Provisioner) UnFormat() (bool, error) {
+	logrus.Infof("%s unformatting Longhorn block device %s", p.name, p.device.Name)
+	return false, nil
+}
+
+func (p *LonghornV1Provisioner) updateDeviceMount(device *diskv1.BlockDevice, devPath string, filesystem *block.FileSystemInfo, needMountUpdate NeedMountUpdateOP) error {
+	logrus.Infof("Prepare to try %s", convertMountStr(needMountUpdate))
+	if device.Status.DeviceStatus.Partitioned {
+		return fmt.Errorf("partitioned device is not supported, please use raw block device instead")
```
votdev, line 291:
```suggestion
		return fmt.Errorf("Partitioned device is not supported, please use raw block device instead")
```

pkg/provisioner/longhornv1.go
```diff
@@ -0,0 +1,457 @@
...
+	if !p.semaphoreObj.acquire() {
+		logrus.Infof("Hit maximum concurrent count. Requeue device %s", device.Name)
+		return true, nil
+	}
+
+	defer p.semaphoreObj.release()
+
+	// before format, we need to unmount the device if it is mounted
+	if filesystem != nil && filesystem.MountPoint != "" {
+		logrus.Infof("unmount %s for %s", filesystem.MountPoint, device.Name)
```
votdev, line 338:
```suggestion
		logrus.Infof("Unmount %s for %s", filesystem.MountPoint, device.Name)
```

pkg/provisioner/longhornv1.go
```diff
@@ -0,0 +1,457 @@
...
+		if err := utils.UmountDisk(filesystem.MountPoint); err != nil {
+			return false, err
+		}
+	}
+
+	// ***TODO***: we should let people to use ext4 or xfs, but now...
+	// make ext4 filesystem format of the partition disk
+	logrus.Debugf("make ext4 filesystem format of device %s", device.Name)
```
votdev, line 346: I don't want to appear small-minded. I think the beginning of a sentence should start with a capital letter.
This should be adapted for all code changes. I will therefore spare myself any further comments in this regard.
```suggestion
	logrus.Debugf("Make ext4 filesystem format of device %s", device.Name)
```

votdev (changes requested):

bk201 (approved):

pkg/provisioner/longhornv1.go
tserong: Actually, because this is an error string it shouldn't be capitalised (see https://go.dev/wiki/CodeReviewComments#error-strings)

pkg/provisioner/longhornv1.go
tserong: I suspect this is form is probably habit related to error strings being uncapitalized. In this case though it's a log message (not an error string) so I agree it should be a proper sentence.

Vicente-Cheng: Thanks, I correct the above logs. For others, we could do it on another PR.

pkg/provisioner/longhornv1.go
Vicente-Cheng: Thanks @tserong. We will keep this for the error log.

pkg/provisioner/longhornv1.go
Vicente-Cheng: done

pkg/provisioner/longhornv1.go
Vicente-Cheng: done

pkg/provisioner/common.go
Vicente-Cheng: I added the related info link on WWN and filesystem UUID. 
Partition is departed. We will remove that in the next version.

pkg/provisioner/common.go
Vicente-Cheng: done

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_blockdevices.yaml
```diff
@@ -84,8 +87,36 @@ spec:
...
                 description: name of the node to which the block device is attached
                 type: string
+              provisioner:
+                properties:
+                  longhorn:
+                    description: a provisioner for provision Longhorn volume backend
+                      disk
+                    properties:
+                      engineVersion:
+                        description: a string with the volume name for the provisioner
```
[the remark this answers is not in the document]
Vicente-Cheng: Nice catch! updated!

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_blockdevices.yaml
```diff
@@ -84,8 +87,36 @@ spec:
...
+                      vgName:
+                        description: a string with the volume group name for the provisioner
+                        type: string
+                    required:
+                    - vgName
+                    type: object
+                type: object
               tags:
-                description: a string with for device tag for provisioner, e.g. "default,small,ssd"
+                description: a string list with device tag for provisioner, e.g. ["default",
```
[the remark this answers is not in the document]
Vicente-Cheng: updated!

pkg/apis/harvesterhci.io/v1beta1/types.go
Vicente-Cheng: updated!

pkg/provisioner/longhornv1.go
votdev: Sorry, missed to remove that comment because of the reason mentioned by Tim.

votdev (approved):
LGTM

Vicente-Cheng:
@mergifyio backport v0.7.x

votdev:
I have a question. We have the scenario that certain labels, e.g. “topology.kubernetes.io/zone” for host resources, are automatically synchronized to Longhorn. Is this scenario considered in this PR and the interface design?

PR #114, opened 2024-06-20
ci: improve the timing of tear down

Vicente-Cheng:
**Problem:**
We destroy VMs no matter whether we deploy VMs or not.
Sometimes, that might break a running test.

Also, I merge two jobs (image build/integration test) together. That can ensure we use the image ASAP

**Solution:**
Only destroy VM if we pass deployment stage

**Related Issue:**
None

**Test plan:**
No need

### DIFF
--- a/.github/workflows/basic-ci.yaml
+++ b/.github/workflows/basic-ci.yaml
@@ -24,25 +24,22 @@ jobs:
         run: |
           make validate
           make validate-ci
-  build:
+  main_jobs:
     needs: validation
     runs-on:
       - self-hosted
       - golang
     steps:
-      - name: "Local Build for next stage"
+      - name: "Clone and check"
+        uses: actions/checkout@v3
+      - name: "Build the Image for the Integration Test"
         run: |
           BUILD_FOR_CI=true make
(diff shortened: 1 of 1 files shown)


.github/workflows/basic-ci.yaml
```diff
@@ -80,11 +77,11 @@ jobs:
           echo Running integration tests
           NDM_HOME=`pwd` go test -v ./tests/...
       - name: "Get NDM logs"
-        if: always()
+        if: steps.vm_deploy.outcome == 'success'
```
tserong, line 80: Could `get-debug-info.sh` still return useful information even if vm_deploy fails? If so, I'd suggest leaving this one as-is, so it always runs.

tserong (approved):
LGTM in general, just one suggestion.

bk201 (approved):

.github/workflows/basic-ci.yaml
Vicente-Cheng: Thanks for the review!
The `get-debug-info.sh` only gets the NDM logs and `blockdevices` CRD.
So, if the node is not deployed, this script also fails.

PR #128, opened 2024-09-05
Add lvm provisioner

Vicente-Cheng:
**Problem:**
Add LVM provisioner

**Solution:**
Add LVM provisioner

**Related Issue:**
https://github.com/harvester/harvester/issues/6014
https://github.com/harvester/harvester/issues/5724

**Test plan:**
TBD


Vicente-Cheng:
codeFactor complains about the generated code. Please ignore that.

pkg/controller/volumegroup/controller.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	vgConds := diskv1.VolumeGroupCondition{
+		Type:               diskv1.VGConditionReady,
+		Status:             corev1.ConditionTrue,
+		LastTransitionTime: metav1.Now(),
+		Reason:             "Volume Group is Ready",
+		Message:            fmt.Sprintf("Volume Group is Ready with devices %v", lvmVG.Spec.Devices),
+	}
+	newConds := UpdateNetworkFSConds(lvmVGCpy.Status.VGConds, vgConds)
+	lvmVGCpy.Status.VGConds = newConds
```
bk201, line 130: might be a wrong naming? I can't figure out the network FS relation here...

pkg/controller/volumegroup/controller.go
```diff
@@ -0,0 +1,259 @@
...
+		lvmVGCpy.Status = &diskv1.VolumeGroupStatus{}
+	}
+	// update Devices:
+	// 1. Spec.Devices >= Status.Devices -> Add/Create PV and VG
+	// 2. Spec.Devices < Status.Devices -> Remove/Reduce PV and VG
+	if lvmVG.Status != nil && len(lvmVG.Status.Devices) == 0 {
+		logrus.Warnf("No devices found in LVMVolumeGroup %s, skip", lvmVG.Name)
+		return nil, nil
+	} else if len(lvmVG.Spec.Devices) >= len(currentDevs) {
+		// add/create pv/vg
```
bk201, line 108: Here the array length is compared. Do we need to consider the situation like changing from `['sda', 'sdb']` to `['sdc', 'sdd']` or `['sde']`.

pkg/provisioner/lvm.go
```diff
@@ -0,0 +1,210 @@
...
+	return false, nil
+}
+
+// Provision creates (if needed) a LVMVg CRD and update the corresponding fields.
+func (l *LVMProvisioner) Provision() (bool, error) {
+	logrus.Infof("Provisioning block device %s to vg: %s", l.device.Name, l.vgName)
+	notFound := false
+
+	// check if the LVMVg CRD is already provisioned
+	targetName := fmt.Sprintf(l.nodeName + "-" + l.vgName)
```
bk201, line 59: Suggest using a non-conflict-generated name here.
We ran into an issue in the backup that the `<namespace>-<name>` in different namespaces might collide before.
For example, creating vg `test-vg` on node `node1` might collide with creating vg `vg` on node `node1-test`. In both cases, they will generate `node1-test-vg` as the VG name.

pkg/provisioner/lvm.go
```diff
@@ -0,0 +1,210 @@
...
+			return true, fmt.Errorf("failed to activate volume group %s, err: %v", l.vgName, err)
+		}
+	} else if lvmvg.Spec.DesireState == diskv1.VGStateDisabled {
+		// make sure the volume group is inactive
+		logrus.Infof("Should not go here, because the LVMVolumeGroup %s should not be disabled", l.vgName)
+	}
+	return
+}
+
+func (l *LVMProvisioner) addDevOrCreateLVMCgCRD(lvmVG *diskv1.LVMVolumeGroup, notFound bool) (requeue bool, err error) {
```
bk201, line 136:
```suggestion
func (l *LVMProvisioner) addDevOrCreateLVMVgCRD(lvmVG *diskv1.LVMVolumeGroup, notFound bool) (requeue bool, err error) {
```

Nit. might need a rename too.

It would be nice to be consistent. The terms "LVMVolumeGroup", "LVMVG", and "LVMVg" are used in the log messages.

bk201 (commented):
Thanks, overall works well, just some suggestions.

pkg/apis/harvesterhci.io/v1beta1/lvm.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// VGConditionReady means the volume group is ready
+	VGConditionReady ConditionType = "Ready"
+	// VGConditionAddDevice means the volume group is added a device
+	VGConditionAddDevice ConditionType = "AddDevice"
+
+	// ConditionTypeReady indicates the volume group is active
+	ConditionTypeActive ConditionType = "Active"
+	// ConditionTypeNotReady indicates the volume group is inactive
+	ConditionTypeInactive ConditionType = "Inactive"
```
bk201, line 36:
```suggestion
	// ConditionTypeActive indicates the volume group is active
	ConditionTypeActive ConditionType = "Active"
	// ConditionTypeInactive indicates the volume group is inactive
	ConditionTypeInactive ConditionType = "Inactive"
```

Just to sync the comment.

pkg/apis/harvesterhci.io/v1beta1/blockdevice.go
```diff
@@ -44,6 +44,10 @@ type BlockDeviceSpec struct {
 	Tags []string `json:"tags,omitempty"`
 
 	Provisioner *ProvisionerInfo `json:"provisioner,omitempty"`
+
+	// a bool for the device to be provisioned
+	// +kubebuilder:default:=false
+	Provisioned bool `json:"provisioned,omitempty"`
```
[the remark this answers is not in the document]
votdev: But I see that it was already called that in several places in the code, so renaming it would only complicate things. 
But the original question still remains whether this field should be correctly named in the present tense.

pkg/apis/harvesterhci.io/v1beta1/lvm.go
```diff
@@ -0,0 +1,119 @@
...
+
+type VolumeGroupStatus struct {
+
+	// The conditions of the volume group
+	// +kubebuilder:validation:Optional
+	// +Kubebuilder:default={}
+	VGConds []VolumeGroupCondition `json:"conditions,omitempty"`
+
+	// The devices of the volume group
+	// format: map[<bd Name>]=devPath"
```
votdev, line 95: Could you add some real life examples here?
Does this field contain data like `map['sda']='/dev/sda'` for example?

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_lvmvolumegroups.yaml
```diff
@@ -0,0 +1,133 @@
...
+              Servers may infer this from the endpoint the client submits requests to.
+              Cannot be updated.
+              In CamelCase.
+              More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
+            type: string
+          metadata:
+            type: object
+          spec:
+            properties:
+              desireState:
```
votdev, line 53: I think `desiredState` is a more correct english. @tserong What do you say to this as a native English speaker?
```suggestion
              desiredState:
```

pkg/controller/volumegroup/controller.go
votdev: The VM import controller has a `ImportConditions` field for example. So maybe change to this in the `VolumeGroupStatus` struct.
```suggestion
	lvmVGCpy.Status.VGConditions = newConds
```

pkg/lvm/lvm.go
```diff
@@ -0,0 +1,84 @@
...
+	ns := utils.GetHostNamespacePath(utils.HostProcPath)
+	executor, err := utils.NewExecutorWithNS(ns)
+	if err != nil {
+		return nil, fmt.Errorf("generate executor failed. %v", err)
+	}
+
+	args := []string{"--noheadings", "-o", "pv_name,vg_name"}
+	output, err := executor.Execute("pvs", args)
+	if err != nil {
+		return nil, fmt.Errorf("pvs failed. %v", err)
```
votdev, line 20:
```suggestion
		return nil, fmt.Errorf("failed to execute 'pvs' command: %v", err)
```

pkg/lvm/lvm.go
```diff
@@ -0,0 +1,84 @@
...
+		pvScanResult[pv] = vg
+	}
+	return pvScanResult, nil
+}
+
+func executeCommandWithNS(cmd string, args []string) error {
+	ns := utils.GetHostNamespacePath(utils.HostProcPath)
+	executor, err := utils.NewExecutorWithNS(ns)
+	if err != nil {
+		return fmt.Errorf("generate executor failed. %v", err)
```
votdev, line 44:
```suggestion
		return fmt.Errorf("generate executor failed: %v", err)
```

pkg/lvm/lvm.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	_, err = executor.Execute(cmd, args)
+	if err != nil {
+		return fmt.Errorf("execute command %s, args: %v failed. %v", cmd, args, err)
```
votdev, line 49:
```suggestion
		return fmt.Errorf("execute command '%s' with args '%v' failed: %v", cmd, args, err)
```

pkg/controller/volumegroup/controller.go
Vicente-Cheng: good catch, rename to `UpdateLVMVGsConds`

pkg/controller/volumegroup/controller.go
Vicente-Cheng: I am not sure about this situation. I rewrote this logic to find the toAdd/toRemove first and then perform the corresponding operations, thinking that it would be more stable.

pkg/provisioner/lvm.go
Vicente-Cheng: yeah, as you mentioned.
I added a specific character for the node name and VG name. It should be unique only when we have the same node and VG names.

pkg/provisioner/lvm.go
Vicente-Cheng: Align with `LVMVolumeGroup` for the log messages.

pkg/apis/harvesterhci.io/v1beta1/lvm.go
Vicente-Cheng: Thanks, updated!

pkg/apis/harvesterhci.io/v1beta1/blockdevice.go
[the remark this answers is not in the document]
Vicente-Cheng: Thanks, I thought we could move to `Porvision`.
This is a new field to replace `Spec.Filesystem.Provisioned`.

Updated!

pkg/apis/harvesterhci.io/v1beta1/lvm.go
Vicente-Cheng: updated!

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_lvmvolumegroups.yaml
Vicente-Cheng: I thought you are correct, updated!

pkg/lvm/lvm.go
Vicente-Cheng: updated!

pkg/lvm/lvm.go
Vicente-Cheng: updated!

pkg/lvm/lvm.go
Vicente-Cheng: updated!

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_lvmvolumegroups.yaml
tserong: Yep, agreed

tserong: It looks like this has been updated to `DesiredState` in the description, but the field names are still `desireState`

pkg/controller/volumegroup/controller.go
```diff
@@ -0,0 +1,263 @@
...
+
+func checkPVAndVG(pvsResult map[string]string, targetPV, targetVG string) (pvFound, vgFound bool, pvCount int) {
+	pvCount = 0
+	for pv, vg := range pvsResult {
+		if pv == targetPV {
+			pvFound = true
+			if vg == targetVG {
+				pvCount++
+				vgFound = true
+				return
```
tserong, line 166: I don't think this return should be here. It means the first time we get a match of both `targetPV` and `targetVG`, we won't count any other PVs that might exist in `targetVG`, which will potentially make the device removal code in `updatePVAndVG()` do the wrong thing.

pkg/controller/volumegroup/controller.go
```diff
@@ -0,0 +1,263 @@
...
+		if !pvFound {
+			if err := lvm.DoPVCreate(dev); err != nil {
+				return err
+			}
+		}
+		if !vgFound {
+			if err := lvm.DoVGCreate(dev, vgCpy.Spec.VgName); err != nil {
+				return err
+			}
+		} else if !pvFound {
```
tserong, line 212: The logic of this `else if !pvFound` line is hurting my brain :-)

From reading the code, it looks like we're doing the following:

- If there's no PV, create one
- If there's no VG, create one from the PV we either found, or the PV we just created
- If there is already a VG and a PV was newly created this time, add the new PV to the VG.
- If there's already both a PV and a VG, do nothing.

Did I get that right?

tserong (commented):
Looks good in general, I just noticed a couple of things

pkg/provisioner/lvm.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	logrus.Infof("Updated LVMVolumeGroup %s, content: %v", l.vgName, lvmVGCpy)
+	return
+}
+
+// genLVMVGName generates the name of the LVMVolumeGroup CRD
+func genLVMVGName(nodeName, vgName string) string {
+	//format N+nodeName+-+V+vgName
+	nHex := hex.EncodeToString([]byte("N"))
+	vHex := hex.EncodeToString([]byte("V"))
```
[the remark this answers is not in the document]
Vicente-Cheng: I would like to make LVMVGName unique, so I would like to have upper `N` and `V` for the start with nodeName and VGName.

But we cannot use any upper for the CRD name, so I encoded it.

pkg/controller/volumegroup/controller.go
Vicente-Cheng: yeah, you are correct.
If we mixed the add/remove together, we would not return at this moment. Let me fix it.

pkg/controller/volumegroup/controller.go
Vicente-Cheng: sure, let me rewrite this logic here.

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_lvmvolumegroups.yaml
Vicente-Cheng: yaml field will start with lower case. You can also check the following nodeName. It's the same case.

pkg/controller/volumegroup/controller.go
tserong: Thanks, I much prefer the new version :-)

pkg/provisioner/lvm.go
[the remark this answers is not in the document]
tserong: Ah, right. I guess without this it would be possible to have collisions if you got very unlucky with node and vg names.  Here's a silly example:

```
NODE NAME         VG NAME        LVMVGNAME
harvester-node-0  test-vg    ->  harvester-node-0-test-vg
harvester-node    0-test-vg  ->  harvester-node-0-test-vg
```

For me though, the encoded `4e` and `56` look weird. Would it be OK to use a lower `n` and `v` instead?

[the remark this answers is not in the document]
tserong: (oops, I see @bk201 already gave a similar example in a separate resolved comment -- sorry for the noise)

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_lvmvolumegroups.yaml
tserong: Sorry, I mean the field names say `desire` as opposed to `desired` (trailing `d`)

pkg/provisioner/lvm.go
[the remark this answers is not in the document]
Vicente-Cheng: Maybe `n-`/`v-` are OK to me. That could also make nodeName/vgName unique.

@bk201, WDYT?

deploy/charts/harvester-node-disk-manager/templates/crds/harvesterhci.io_lvmvolumegroups.yaml
Vicente-Cheng: nice catch... I missed that one

Vicente-Cheng: updated!

pkg/provisioner/lvm.go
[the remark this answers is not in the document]
Vicente-Cheng: I use the new mechanism.
Use the naming prefix `vgName-` with `GenerateName`.
That means k8s will give the random string for the full name like `vgName-xxxx`.
We also added a topology key value (node Name) to the label.

So, to find the target lvmvg CR, the selector (by the above label) will be used.

That would resolve the naming conflict issue.

Please also recheck it. Thanks!

pkg/provisioner/lvm.go
```diff
@@ -0,0 +1,231 @@
...
+		return
+	}
+	for _, lvmvg := range lvmvgs.Items {
+		if lvmvg.Spec.NodeName == l.nodeName && lvmvg.Spec.VgName == l.vgName {
+			target = &lvmvg
+			found = true
+			break
+		}
+	}
+	return
```
bk201, line 230: By returning a NotFound error we can eliminate the `found` variable. And callers can check the error.

pkg/provisioner/lvm.go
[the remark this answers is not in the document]
tserong: Neat!

pkg/provisioner/lvm.go
tserong: Good idea

Vicente-Cheng: updated!

pkg/provisioner/lvm.go
```diff
@@ -0,0 +1,240 @@
+package provisioner
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/pkg/errors"
```
tserong, line 7: I think this should just be "errors" (i.e. without the "github.com/pkg" bit at the start)?

pkg/provisioner/lvm.go
```diff
@@ -0,0 +1,240 @@
...
+
+	diskv1 "github.com/harvester/node-disk-manager/pkg/apis/harvesterhci.io/v1beta1"
+	"github.com/harvester/node-disk-manager/pkg/block"
+	ctldiskv1 "github.com/harvester/node-disk-manager/pkg/generated/controllers/harvesterhci.io/v1beta1"
+	"github.com/harvester/node-disk-manager/pkg/lvm"
+	"github.com/harvester/node-disk-manager/pkg/utils"
+)
+
+var (
+	NotFoundError = errors.New("LVMVG CR is not found")
```
tserong, line 19: The validation check is complaining that this should be called `ErrNotFound`

pkg/provisioner/lvm.go
```diff
@@ -0,0 +1,240 @@
...
+		return
+	}
+	lvmvgs, err := l.vgClient.List(utils.HarvesterNS, metav1.ListOptions{LabelSelector: selector.String()})
+	if err != nil {
+		err = fmt.Errorf("failed to list LVMVolumeGroup %s. %v", l.vgName, err)
+		return
+	}
+	for _, lvmvg := range lvmvgs.Items {
+		if lvmvg.Spec.NodeName == l.nodeName && lvmvg.Spec.VgName == l.vgName {
+			target = &lvmvg
```
tserong, line 231: This is flagging `G601: Implicit memory aliasing in for loop. (gosec)`, but that might be a false positive for us now (see https://stackoverflow.com/a/68247837)

tserong (commented):
I did a build of this and gave it a quick test on a single node cluster with two extra disks, and it works nicely (I provisioned and unprovisioned both disks to the same VG and it looks like it's behaving correctly).

Regarding the NotFoundError, it might be preferable to import "k8s.io/apimachinery/pkg/api/errors" then call `errors.NewNotFound(...)` to create the error, and `errors.IsNotFound()` to check it, rather than having our own error variable. If that makes sense, you can ignore my comments below regarding error variable naming and import paths :-)

pkg/provisioner/lvm.go
Vicente-Cheng: I use the `k8s.io/apimachinery/pkg/api/errors` to fit the suggestion. Updated.

pkg/provisioner/lvm.go
Vicente-Cheng: ditto

bk201 (approved):
LGTM

tserong (approved):
LGTM!

tserong:
@mergifyio backport v0.7.x

PR #129, opened 2024-09-09
feat: add Longhorn V2 provisioner

tserong:
**Problem:**
We need to add a provisioner for LH V2 disks

**Solution:**
This PR

**Related Issue:**
https://github.com/harvester/harvester/issues/6014

**Test plan:**
TBD

**Note:**
The first two commits in this PR are cherry picks from https://github.com/harvester/node-disk-manager/pull/128, because I needed the changes to the blockdevices CRD. I'll drop/rebase appropriately once that PR goes in.


tserong:
OK, that force push should have fixed the dirty validation, and we can ignore CodeFactor, because it's just whining about generated code :-)

Some further comments lifted from the commit message on the last commit:

I've implemented this as a separate `LonghornV2Provisioner` for the sake of neatness, rather than grafting it into `LonghornV1Provisioner` with a bunch of `if` statements, but some of the implementation (the `UnProvision()` and `Update()` functions) still call back to the `LonghornV1Provisioner` to avoid excessive code duplication.

The current implementation theoretically supports using virtio disks, except that NDM as a whole ignores those, as they don't have WWNs. This is something we need to address separately as part of https://github.com/harvester/harvester/issues/6034

bk201 (approved):

Vicente-Cheng (approved):
LGTM, but we need to wait for the LVM provisioner and rebase again.

tserong:
@mergifyio backport v0.7.x

tserong:
https://github.com/Mergifyio backport v0.7.x

tserong:
OK, I give up, obviously I can't get mergify to redo a backport PR, I'm just going to do this by hand...

PR #130, opened 2024-09-11
Use BusPath to resolve dm device paths

tserong:
**Problem:**
If multipathd is running and has taken over a device, `ResolvePersistentDevPath()` will end up returning a `/dev/dm-*` device path, which we don't want.  We want the real underyling device (e.g. `/dev/sda`).  If we take a `/dev/dm-*` path and later update the blockdevice CR with it, we lose all the interesting DeviceStatus information, like the WWN.

Happily, in this case, we can figure out the right path to use by checking `/dev/disk/by-path/` + the device's bus path. This has the added advantage of also working for block devices that have no WWN.

**Solution:**
This PR.

Note: we're not trying to make NDM support multipath devices here, we're just trying to avoid having it corrupt the blockdevice CRs if multipath happens to be active.

**Related Issue:**
https://github.com/harvester/harvester/issues/6531

**Test plan:**
- Add an extra disk that has a WWN (e.g. a scsi disk) to a harvester node
- Start multipathd (this may require editing the grub command line to remove `multipath=off`)
- Verify that the disk has been taken by multipathd:
```
# multipath -l
35000c50015ac3bd9 dm-0 QEMU,QEMU HARDDISK
size=20G features='0' hwhandler='0' wp=rw
`-+- policy='service-time 0' prio=0 status=active
  `- 2:0:0:0 sda 8:0 active undef running
```
- Use the Harvester GUI to add the disk to the host, as described in https://github.com/harvester/harvester/issues/6531.
- Note that it should now _correctly_ show up as `/dev/sda`, not as `/dev/dm-0` as in that bug report.
- At this point, force formatting will fail, because the device has been taken by multipath. You will see an error like the following. That's fine.
image
- Run `multipath -F` to flush the multipath config. This will remove the device from multipath. It should be automatically formatted shortly afterwards.

### DIFF
--- a/pkg/controller/blockdevice/controller.go
+++ b/pkg/controller/blockdevice/controller.go
@@ -137,9 +137,6 @@ func (c *Controller) OnBlockDeviceChange(_ string, device *diskv1.BlockDevice) (
 	if err != nil {
 		return nil, err
 	}
-	if devPath == "" {
-		return nil, fmt.Errorf("failed to resolve persistent dev path for block device %s", device.Name)
-	}
 
 	logrus.Debugf("Checking to format device %s", device.Name)
 	if formatted, requeue, err := provisionerInst.Format(devPath); !formatted {
--- a/pkg/provisioner/common.go
+++ b/pkg/provisioner/common.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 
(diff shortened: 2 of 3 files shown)


Vicente-Cheng (approved):
LGTM, please rebase to use the latest CI

tserong:
Rebased

Vicente-Cheng:
@mergifyio backport v0.7.x

bk201 (approved):
thanks!

PR #136, opened 2024-09-21
controller/scanner: should consider the wwn of the added blockdevice

Vicente-Cheng:
- We should also consider the wwn of the added blockdevice in
      the same round.

**Problem:**
It would not work well if two duplicate WWN devices wanted to join simultaneously.

**Solution:**
Should consider the WWNs even if in the same round

**Related Issue:**
https://github.com/harvester/harvester/issues/6604

**Test plan:**


tserong (approved):

Vicente-Cheng:
@mergifyio backport v0.7.x

bk201 (approved):

PR #139, opened 2024-09-24
controller: consider the new provision field

Vicente-Cheng:
**Problem:**
Missing mountpoint when provisioning the longhorn V1

**Solution:**
We need to improve the handling with the new provision field `Spec.Provision`.
Walk through the whole `Spec.Filesystem.Provisioned` logic and improve them.

**Related Issue:**
None

**Test plan:**
Make sure the longhornV1 can works as usual only with the new provision field `Spec.Provision`.


Vicente-Cheng:
@mergifyio backport v0.7.x

bk201 (approved):

tserong (approved):

PR #142, opened 2024-09-26
Fix two issues encountered when testing LHv2 provisioning

tserong:
**Problem:**
1. Disks added for use by LHv2 are added successfully, but their status remains stuck with an error "spec and status of disks on node $NAME are being syncing and please retry later."
2. Active LHv2 volumes attached to VMs appear as `/dev/dm-*` and `/dev/nvme*` devices on the host. NDM sees these and thinks they're real disks, and then creates new BD CRs from those devices.

**Solution:**
1. Add a call to `reflect.DeepEqual()` to avoid unnecessary updates and resultant weird state.
2. Abuse the existing vendor filter to pick up SPDK devices, and explicitly exclude `/dev/dm-` devices in `ApplyExcludeFiltersForDisk()`.

Note: I intend to handle this more cleanly later as part of https://github.com/harvester/harvester/issues/5059

**Related Issue:**

**Test plan:**
N/A

### DIFF
--- a/pkg/filter/vendor_filter.go
+++ b/pkg/filter/vendor_filter.go
@@ -8,10 +8,14 @@ import (
 const (
 	vendorFilterName            = "vendor filter"
 	vendorFilterDefaultLonghorn = "longhorn"
+	// This is a hack to make sure we skip active longhorn v2 volumes,
+	// which appear as /dev/nvme* devices.  They don't have a vendor,
+	// but do specify ID_MODEL=SPDK bdev Controller...
+	modelFilterDefaultSPDK = "SPDK bdev Controller"
 )
 
 var (
-	defaultExcludedVendors = []string{vendorFilterDefaultLonghorn}
+	defaultExcludedVendors = []string{vendorFilterDefaultLonghorn, modelFilterDefaultSPDK}
 )
 
 type vendorFilter struct {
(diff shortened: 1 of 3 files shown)


pkg/filter/vendor_filter.go
```diff
@@ -8,10 +8,14 @@ import (
...
 	vendorFilterDefaultLonghorn = "longhorn"
+	// This is a hack to make sure we skip active longhorn v2 volumes,
+	// which appear as /dev/nvme* devices.  They don't have a vendor,
+	// but do specify ID_MODEL=SPDK bdev Controller...
+	modelFilterDefaultSPDK = "SPDK bdev Controller"
 )
 
 var (
-	defaultExcludedVendors = []string{vendorFilterDefaultLonghorn}
+	defaultExcludedVendors = []string{vendorFilterDefaultLonghorn, modelFilterDefaultSPDK}
```
Vicente-Cheng, line 18: Would you like to change the name of this variable?
Now, we mixed the vendor/module into this slice.

I do not have a strong opinion about changing the name. If you want to keep it, I am also okay with that.
Thanks!

Vicente-Cheng, line 18: Would you like to change the name of this variable?
Now, we mixed the vendor/module into this slice.

I do not have a strong opinion about changing the name. If you want to keep it, I am also okay with that.
Thanks!

Vicente-Cheng (approved):
just a nit, others LGTM!

Thanks for the enhancement.

bk201 (approved):

Vicente-Cheng:
@mergifyio backport v0.7.x

Vicente-Cheng:
Failure is the side-effect that comes from https://github.com/harvester/node-disk-manager/blob/master/pkg/controller/blockdevice/scanner.go#L287

The autoProvision should not update both `Spec.Filesystem.Provisioned` and `Spec.Provision`.
If these two fields were true, the remove operation would fail.

pkg/filter/vendor_filter.go
tserong: I'd like to keep it for now - I know it's a bit silly/incorrect, I just didn't rename it given that I want to rework all this filter stuff anyway separately for https://github.com/harvester/harvester/issues/5059

PR #143, opened 2024-09-26
ci: make sure the if-condition works

Vicente-Cheng:
**Problem:**
The CI teardown flow is not working.

Need always() to trigger the steps, or steps after the failure cannot be executed.

**Solution:**
Make sure the if-condition works

**Related Issue:**

**Test plan:**

### DIFF
--- a/.github/workflows/basic-ci.yaml
+++ b/.github/workflows/basic-ci.yaml
@@ -11,6 +11,7 @@ on:
 
 env:
   LIBVIRT_DEFAULT_URI: "qemu:///system"
+  VM_DEPLOYED: false
 
 jobs:
   validation:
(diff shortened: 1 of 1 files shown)


bk201 (approved):

tserong (approved):
LGTM, although I wonder -- could we instead _remove_ the `if` line entirely? Would that also work to make these steps run?

Vicente-Cheng:
> LGTM, although I wonder -- could we instead _remove_ the `if` line entirely? Would that also work to make these steps run?

Sure, that's the original behavior.
With always() but no check.

So I would get some error log from getting NDM log or vagrant destroy
I would like to cleanup these errors.

Vicente-Cheng:
@mergifyio backport v0.7.x v0.6.x

PR #150, opened 2024-09-30
Add fresh installation CI integration job

Vicente-Cheng:
**Problem:**
We should have integration tests for both upgrade/fresh installation.

**Solution:**
Add another job for fresh installation.

**Related Issue:**

**Test plan:**

### DIFF
--- a/.github/workflows/basic-ci.yaml
+++ b/.github/workflows/basic-ci.yaml
@@ -25,7 +25,7 @@ jobs:
         run: |
           make validate
           make validate-ci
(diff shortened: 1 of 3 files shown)


bk201 (approved):

Vicente-Cheng:
The upgrade job failure should be handled when we introduce the mutator.
So I need this for the further development. (I mean, we could check the whole thing from both upgrade/fresh installation parts)

Vicente-Cheng:
close due to #151

PR #151, opened 2024-10-06
Add webhook

Vicente-Cheng:
**Problem:**
We need webhook to check and patch the blockdevice CR.

**Solution:**
Add webhook

**Related Issue:**

**Test plan:**


Vicente-Cheng:
Hi @tserong, @bk201

Please review this PR. I use the new CI implementation (from PR: https://github.com/harvester/node-disk-manager/pull/150)
to verify both upgrade/fresh installation scenarios. (If we merge this PR, we could close the above PR)

Also, we need this PR https://github.com/harvester/webhook/pull/7 so I can update the vendor again.

Please review the main logic about webhook and drop `spec.filesystem.provisioned`.
Thanks!

cmd/node-disk-manager-webhook/main.go
```diff
@@ -0,0 +1,160 @@
...
+			Usage:       "The harvester namespace",
+			Value:       "harvester-system",
+			Required:    true,
+		},
+		&cli.StringFlag{
+			Name:        "controller-user",
+			EnvVars:     []string{"CONTROLLER_USER_NAME"},
+			Destination: &options.ControllerUsername,
+			Value:       "harvester-node-disk-manager-webhook",
+			Usage:       "The harvester controller username",
```
bk201, line 68:
```suggestion
			Usage:       "The harvester-node-disk-manager controller username",
```

package/Dockerfile.webhook
```diff
@@ -0,0 +1,22 @@
+# syntax=docker/dockerfile:1.7.0
+
+FROM registry.suse.com/bci/bci-base:15.6
+
+# util-linux-systemd -> for `lsblk` command
```
bk201, line 5: suggest removing unused codes

pkg/webhook/blockdevice/mutator.go
```diff
@@ -0,0 +1,88 @@
...
+	BlockdeviceCache ctldiskv1.BlockDeviceCache
+}
+
+func NewBlockdeviceMutator(blockdeviceCache ctldiskv1.BlockDeviceCache) *Mutator {
+	return &Mutator{
+		BlockdeviceCache: blockdeviceCache,
+	}
+}
+
+func (m *Mutator) Update(_ *admission.Request, oldObj, newObj runtime.Object) (admission.Patch, error) {
```
bk201, line 24: can you add a comment block to briefly describe why we need to mutate the resource (a link to HEP is also OK)?

Vicente-Cheng: updated!

package/Dockerfile.webhook
Vicente-Cheng: nice catch... removed

cmd/node-disk-manager-webhook/main.go
Vicente-Cheng: Updated

pkg/webhook/blockdevice/mutator.go
```diff
@@ -0,0 +1,93 @@
...
+
+	if !prevProvision && newBd.Spec.FileSystem.Provisioned {
+		provisionVal := newBd.Spec.FileSystem.Provisioned
+		patchAddNewProvision := admission.PatchOp{
+			Op:    admission.PatchOpReplace,
+			Path:  "/spec/provision",
+			Value: provisionVal,
+		}
+		patchOps = append(patchOps, patchAddNewProvision)
+		if provisionVal && newBd.Spec.Provisioner == nil {
```
tserong, line 52: `provisionVal` will always be true here, so this could just be `if newBd.spec.provisioner == nil`.

(Also, given it's always true, the `provisionVal` variable could be dropped entirely and the above PatchOp could just specify `Value: true` - although I don't have a strong preference here either way)

pkg/webhook/blockdevice/mutator.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	// means we need to disable, align the .spec.filesystem.provisioned with .spec.provision -> false
+	if prevProvision && (!newBd.Spec.Provision || !newBd.Spec.FileSystem.Provisioned) {
+		if newBd.Spec.Provision {
+			patchProvision := admission.PatchOp{
+				Op:    admission.PatchOpReplace,
+				Path:  "/spec/provision",
+				Value: false,
+			}
+			patchOps = append(patchOps, patchProvision)
```
tserong, line 73: Should there also be a patch op here to explicitly remove Spec.FileSystem.Provisioned?

pkg/controller/blockdevice/controller.go
```diff
@@ -207,26 +211,32 @@ func (c *Controller) finalizeBlockDevice(oldBd, newBd *diskv1.BlockDevice, devPa
...
 func (c *Controller) generateProvisioner(device *diskv1.BlockDevice) (provisioner.Provisioner, error) {
-	if device.Spec.Provisioner == nil && !device.Spec.FileSystem.Provisioned && device.Status.ProvisionPhase != diskv1.ProvisionPhaseProvisioned {
+	// skip the non-provisioned device
+	if device.Spec.Provisioner == nil && device.Status.ProvisionPhase != diskv1.ProvisionPhaseProvisioned {
+		return nil, nil
+	}
+	// upgrade case, we need to update some fields
+	if device.Spec.Provisioner == nil && device.Status.ProvisionPhase == diskv1.ProvisionPhaseProvisioned {
+		device.Spec.Provision = true
+		device.Spec.FileSystem.Provisioned = false
```
tserong, line 221: ISTM that by setting `Spec.Filesystem.Provisioned = false` here, the mutator (which I assume runs later when the BD is subsequently updated?) will hit the `prevProvision && (!newBd.Spec.Provision || !newBd.Spec.Filesystem.Provisioned)` case, and will then go and patch in `Spec.Provision = false`, thus unprovisioning the device.

Or am I misunderstanding the order of events?

tserong (commented):
I have a couple of questions :-)

pkg/webhook/blockdevice/mutator.go
Vicente-Cheng: yeah, you are correct. Simplified this part.

pkg/webhook/blockdevice/mutator.go
Vicente-Cheng: No, it seems the admission mutate cannot update the deprecated part.
The controller now only checks the `spec.provision`, so we ignore this.

pkg/controller/blockdevice/controller.go
Vicente-Cheng: Yeah, this might happen. So, the mutator will ignore the changes from the controller.
You can see we have the precheck as below:
```
        if req.IsFromController() {
		return nil, nil
	}
```

After that, we could avoid the case you mentioned.

tserong: Ah! Thank you, I had missed that detail.

tserong (approved):

bk201 (approved):

Vicente-Cheng:
@mergifyio backport v0.7.x

PR #152, opened 2024-10-10
fix: set LHv2 DiskDriver to "auto" if unset

tserong:
**Problem:**
We need to force DiskDriver to "auto" if it's not explicitly set, because Longhorn also does that internally.  If we don't do this, the subsequent `reflect.DeepEqual()` in `LonghornV2Provisioner.Update()` will always fail because we have an empty string, but the LHN CR will have it set to "auto" which results in a weird resync loop.

**Solution:**
This PR.

**Related Issue:**
https://github.com/harvester/harvester/issues/6709

**Test plan:**
Add an LHv2 disk. Verify we don't see the errors mentioned in the related issue.

### DIFF
--- a/pkg/provisioner/longhornv2.go
+++ b/pkg/provisioner/longhornv2.go
@@ -30,6 +30,14 @@ func NewLHV2Provisioner(
 	if !cacheDiskTags.Initialized() {
 		return nil, errors.New(ErrorCacheDiskTagsNotInitialized)
 	}
+	if device.Spec.Provisioner.Longhorn.DiskDriver == longhornv1.DiskDriverNone {
+		// We need to force DiskDriver to "auto" if it's not explicitly set,
+		// because Longhorn also does that internally.  If we don't do it
+		// here, the subsequent reflect.DeepEqual() in our Update() function
+		// will always fail because we have an empty string, but the LHN CR
+		// has it set to "auto" which results in a weird resync loop.
+		device.Spec.Provisioner.Longhorn.DiskDriver = longhornv1.DiskDriverAuto
+	}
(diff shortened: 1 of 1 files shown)


tserong:
https://github.com/bk201/vagrant-k3s/pull/3 should fix that CI failure

Vicente-Cheng (approved):
LGTM, thanks!

bk201 (approved):

tserong:
latest CI failure is:

```
=== RUN   TestSingleDiskOperation/Test_1_UnprovisionSingleDisk
    test_0_single_disk_test.go:141: 
        	Error Trace:	/home/github/actions-runner/_work/node-disk-manager/node-disk-manager/tests/integration/test_0_single_disk_test.go:141
        	Error:      	Not equal: 
        	            	expected: "/var/lib/harvester/extra-disks/9665fd40086d48e7ca541b1e990c6645"
        	            	actual  : ""
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-/var/lib/harvester/extra-disks/9665fd40086d48e7ca541b1e990c6645
        	            	+
        	Test:       	TestSingleDiskOperation/Test_1_UnprovisionSingleDisk
        	Messages:   	Mountpoint should be empty after we remove disk!
```

I suspect this is probably because `Test_1_UnprovisionSingleDisk()` is still setting `Spec.FileSystem.Provisioned = false` rather than `Spec.Provision = false`, and is unrelated to this PR.

Vicente-Cheng:
Yeah, and it should be fixed on #151. I thought we could merge it and let CI re-run with the master branch

PR #155, opened 2024-10-16
Add new validator when adding bd

Vicente-Cheng:
**Problem:**
Because volume group management needs to consider many factors, we decided to offload this behavior to the user and block operations, like adding a new block device to the volume group. 

**Solution:**
We need to check when trying to add a new blockdevice into volume group

**Related Issue:**

**Test plan:**


pkg/webhook/blockdevice/validator.go
```diff
@@ -43,6 +59,57 @@ func (v *Validator) validateProvisioner(bd *diskv1.BlockDevice) error {
...
+	}
+
+	// check if there is any PV created with the targetSC
+	pvs, err := v.pvCache.List(labels.Everything())
+	if err != nil {
+		return werror.NewBadRequest("Failed to list PVs")
+	}
+	for _, pv := range pvs {
+		if pv.Spec.StorageClassName == targetSC {
+			return werror.NewBadRequest("There is already a PVC created with the target volume group, we cannot add any blockdevice into it")
```
tserong, line 107:
```suggestion
			return werror.NewBadRequest("There is already a PVC created using the target volume group, so we cannot add or remove the associated blockdevices")
```

tserong (approved):
Just a very small wording suggestion, otherwise LGTM!

pkg/webhook/blockdevice/validator.go
Vicente-Cheng: updated

tserong (approved):

tserong (approved):

bk201 (approved):

Vicente-Cheng:
@mergifyio backport v0.7.x

PR #157, opened 2024-10-21
fix: Call `wipefs -a` in LonghornV2Provisioner.Format()

tserong:
**Problem:**
When using the Longhorn V2 data engine, for NVMe devices where Longhorn decides to use the nvme bdev driver, device activation will fail if there's an existing filesystem. This does not appear to be a problem when using the aio driver.

**Solution:**
Run `wipefs -a $devPath` during provisioning.

Without this, we'll potentially have devices stuck unschedulable with an obscure error message like this:

`Disk c1fd02d173a3f4b9176705e9b2e37d39(0003:03:00.0) on node altra is not ready: failed to generate disk config: error: rpc error: code = Internal desc = rpc error: code = Internal desc = failed to add disk block device: failed to create disk bdev: failed to attach NVMe disk 0003:03:00.0: error sending message, id 3126, method bdev_nvme_attach_controller, params {c1fd02d173a3f4b9176705e9b2e37d39
{PCIe  0003:03:00.0  }   30 2 15 disable}: {"code": -19,"message":
"No such device"}`

**Related Issue:**
https://github.com/harvester/harvester/issues/6828

**Test plan:**
- On a host with an extra nvme device, run `mkfs.ext4 /dev/nvme0n1` (or whatever the device name is) to make a filesystem.
- Try to add the disk to the host for Longhorn V2 usage. It should be stuck unschedulable (check `kubectl -n longhorn-system get lhn -o yaml` and look under `diskStatus` for an error message similar to the description above).
- Re-test with this fix applied. The disk should provision fine.

### DIFF
--- a/pkg/provisioner/longhornv2.go
+++ b/pkg/provisioner/longhornv2.go
@@ -13,6 +13,7 @@ import (
 	diskv1 "github.com/harvester/node-disk-manager/pkg/apis/harvesterhci.io/v1beta1"
 	"github.com/harvester/node-disk-manager/pkg/block"
 	ctllonghornv1 "github.com/harvester/node-disk-manager/pkg/generated/controllers/longhorn.io/v1beta2"
+	"github.com/harvester/node-disk-manager/pkg/utils"
 )
 
 type LonghornV2Provisioner struct {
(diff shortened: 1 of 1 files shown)


bk201 (approved):

Vicente-Cheng (approved):
thanks

pkg/provisioner/longhornv2.go
```diff
@@ -55,11 +56,17 @@ func NewLHV2Provisioner(
...
-// isFormatComplete == true to indicate the disk is ready for use.
-func (p *LonghornV2Provisioner) Format(string) (isFormatComplete, isRequeueNeeded bool, err error) {
-	isFormatComplete = true
-	return
+// Format should really be a no-op for V2 disks given they just take the
+// whole device, but for NVMe devices where Longhorn decides to use the
+// nvme bdev driver, device activation will fail if there's an existing
+// filesystem on the device, so we need to make sure to wipe before use.
+func (p *LonghornV2Provisioner) Format(devPath string) (isFormatComplete, isRequeueNeeded bool, err error) {
+	cmd := exec.Command("wipefs", "-a", devPath)
```
Vicente-Cheng, line 64: Would you like to use https://github.com/harvester/node-disk-manager/blob/master/pkg/utils/command.go?

The go-common has a similar [one](https://github.com/harvester/go-common/tree/main/command), we could also keep this here. And move to go-common/command after v1.4

Vicente-Cheng (approved):
just a nit

pkg/provisioner/longhornv2.go
tserong: Yeah, that's a bit neater :-) Done.

tserong:
@mergifyio backport v0.7.x

PR #159, opened 2024-10-23
provisioner: ensure the device is ready before adding into volume group

Vicente-Cheng:
- The filesystem metadata would break the pvcreate

**Problem:**
The filesystem metadata would break the pvcreate

**Solution:**
make sure the device is clean

**Related Issue:**

**Test plan:**


Vicente-Cheng:
@mergifyio backport v0.7.x

tserong (approved):

bk201 (approved):

PR #161, opened 2024-10-24
Add webhook to ensure the VGStatus is active

Vicente-Cheng:
**Problem:**
We need to fix the multiple times wipefs, which might break the metadata.
Also, add the webhook to ensure the VGStatus is active.

**Solution:**
add the webhook to ensure the VGStatus is active
ensure we only execute wipefs before pvcreate

**Related Issue:**

**Test plan:**


bk201 (approved):

pkg/provisioner/lvm.go
```diff
@@ -78,6 +83,7 @@ func (l *LVMProvisioner) Provision() (bool, error) {
 	if err != nil {
 		return requeue, err
 	}
+	l.lock.Unlock()
```
tserong, line 86: This should probably be `defer l.lock.Unlock()` and moved to immediately after line 74, else any error above will result in the lock remaining in place.

tserong (changes requested):
One small suggestion, otherwise LGTM.

pkg/provisioner/lvm.go
Vicente-Cheng: nice catch!
I originally wanted to reduce the lock scope, so I did not use defer func (but yes, I missed the error handling)
It looks like it did not have a huge difference with/ lock the whole function, so like you mentioned, defer function would be nice.
Thanks!

tserong (approved):

Vicente-Cheng:
@mergifyio backport v0.7.x

PR #174, opened 2024-12-31
ci: Bump image version in the Harvester repo when tagging

bk201:
Add a workflow to create a PR in the harvester/harvester repo when seeing a tag event. The workflow will:
- Determine the stable branch in this repo. e.g., version `v0.7.8` -> branch `v0.7.x`.
- Reading the file `.github/branch-mappings.yaml` and determine the corresponding Harvester branch. e.g., `v0.7.x` -> `v1.4`
- Clone harvester/harvester repo, checkout corresponding branch, and modify the component's image tags in `deploy/charts/harvester/values.yaml`
- Submit a PR. Note a secret named `HARVESTER_BOT_TOKEN` needs to be configured in the repo for actions. The token in the secret must have permission to create PRs in the harvester/harvester repo.

The file `.github/branch-mappings.yaml` is read from the default branch. When we have a new mapping, we only need to update the file in the default branch.

- sample action: https://github.com/bk201/node-disk-manager/actions/runs/12556466870
- sample PR: https://github.com/bk201/harvester/pull/6

### DIFF
--- a/.github/branch-mappings.yaml
+++ b/.github/branch-mappings.yaml
@@ -1,0 +1,6 @@
+# branch mapping from this repo to harvester
+mappings:
+  v0.5.x: v1.2
+  v0.6.x: v1.3
+  v0.7.x: v1.4
+  v0.8.x: v1.5
--- a/.github/workflows/create-image-bump-pr.yaml
+++ b/.github/workflows/create-image-bump-pr.yaml
@@ -1,0 +1,87 @@
+name: Create PR to bump image tags in the Harvester repo
+
+on:
+  workflow_call:
+    inputs:
+      component:
+        description: 'The component to bump images'
+        required: true
+        type: string
+      version:
+        description: 'The version of the image to bump'
+        required: true
+        type: string
+      harvester_repo:
+        description: 'The repository to bump the image in'
+        default: 'harvester/harvester'
+        required: false
+        type: string
+
+jobs:
+  create:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Dump GitHub context
+        env:
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+        run: echo "$GITHUB_CONTEXT"
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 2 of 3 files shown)


.github/branch-mappings.yaml
```diff
@@ -0,0 +1,6 @@
+# branch mapping from this repo to harvester
```
innobead, line 1: This is good.

bk201: I'm struggling with this. To centralize in one place or controlled in each repo.

innobead: Can we add an action in the harvester repo to regularly scan these component repos, check for any new tags or releases, and update the related tags accordingly?

bk201: The initial idea is repo-based. The intention is just to make the owner's life easier. 
I'll close this one and see what I can.

Vicente-Cheng: How about the submodule to handle the whole components update?
If we centralize this thing, maybe we could use other ways, not just bash-based GHA.
Could we do it easier?

bk201: The reason to keep this inside a repo is the workflow can be run after images are successfully published. Or even run after some tests (if there are any) defined by the repo.
Using a separate job to scan all component repos will be more complicated, we'll need to check if CI is running well when the tag was happening.

How about creating an action, and every repo just needs to use the action?

Vicente-Cheng: I am ok with that.
Just one thing I would like to confirm again. Suppose we use the centralized GitHub action. That means we need to use the same branch naming, right?

bk201: no, a branch_mapping mapping file should be kept in the individual repo (just like this PR). I prefer this rather than keeping it in a centralized place. The advantage is the repo owner can know and update the mapping when working on the repo, no need to jump to the centralized repo. The drawback is we don't have a place to know all the mappings for all projects in one place.

PR #195, opened 2025-05-07
controller/blockdevice: should update DevPath once it changed

Vicente-Cheng:
- when the node reboot, it might cause the device path change because of the interrupt order with the device. We need to update the device path once it changed.

**Problem:**
Sometimes we will see the duplicated status.deviceStatus.devPath of the blockdevice CR in the same node.

The device path would be change in following two cases:
1. device is temporarily gone and back.
2. Node reboot might also cause the device path change (depends on device interrupt).

So, we need to update the device path once it has changed

**Solution:**
Update device path once it has changed

**Related Issue:**
https://github.com/harvester/harvester/issues/8164

**Test plan:**
Try to reboot, and you might see the spec.devPath is different from status.deviceStatus.devPath.
**NOTE**: It is hard to reproduce so that we can focus on the general use case, which should be kept the same.

### DIFF
--- a/pkg/controller/blockdevice/scanner.go
+++ b/pkg/controller/blockdevice/scanner.go
@@ -127,17 +127,18 @@ func (s *Scanner) collectAllDevices() []*deviceWithAutoProvision {
 
 func (s *Scanner) handleExistingDev(oldBd *diskv1.BlockDevice, newBd *diskv1.BlockDevice, autoProvisioned bool) {
 	if isDevPathChanged(oldBd, newBd) {
-		// the Dev Path changed only when the device is temporarily gone and back.
-		if oldBd.Status.State != diskv1.BlockDeviceInactive {
-			logrus.Warnf("Device Path should not change with the active device %s, new device path: %s", newBd.Name, newBd.Status.DeviceStatus.DevPath)
-		} else {
+		// Dev Path will change when device is temporarily gone and back.
+		// Node reboot might also cause the device path change (depends on device interrupt).
+		// According to the above, we should update the device path once it changes.
+		if oldBd.Status.State == diskv1.BlockDeviceInactive {
 			logrus.Infof("The inactive block device %s with wwn %s is coming back", newBd.Name, newBd.Status.DeviceStatus.Details.WWN)
 			oldBd.Status.State = diskv1.BlockDeviceActive
-			oldBd.Status.DeviceStatus.DevPath = newBd.Status.DeviceStatus.DevPath
-			if _, err := s.Blockdevices.Update(oldBd); err != nil {
-				logrus.Errorf("Update device %s status error", oldBd.Name)
-				s.Blockdevices.Enqueue(s.Namespace, newBd.Name)
-			}
+		}
+		logrus.Infof("Try to update the device path of %s, which is changed from %s to %s.", oldBd.Name, oldBd.Status.DeviceStatus.DevPath, newBd.Status.DeviceStatus.DevPath)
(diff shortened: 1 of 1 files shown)


pkg/controller/blockdevice/scanner.go
```diff
@@ -127,17 +127,17 @@ func (s *Scanner) collectAllDevices() []*deviceWithAutoProvision {
...
 			oldBd.Status.State = diskv1.BlockDeviceActive
-			oldBd.Status.DeviceStatus.DevPath = newBd.Status.DeviceStatus.DevPath
-			if _, err := s.Blockdevices.Update(oldBd); err != nil {
-				logrus.Errorf("Update device %s status error", oldBd.Name)
-				s.Blockdevices.Enqueue(s.Namespace, newBd.Name)
-			}
+		}
+		oldBd.Status.DeviceStatus.DevPath = newBd.Status.DeviceStatus.DevPath
+		if _, err := s.Blockdevices.Update(oldBd); err != nil {
+			logrus.Errorf("Update device %s status error, wake up scanner again", oldBd.Name)
```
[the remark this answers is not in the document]
Vicente-Cheng: nice suggestion

Vicente-Cheng:
@mergifyio backport v1.5

pkg/controller/blockdevice/scanner.go
```diff
@@ -127,17 +127,18 @@ func (s *Scanner) collectAllDevices() []*deviceWithAutoProvision {
...
-			if _, err := s.Blockdevices.Update(oldBd); err != nil {
-				logrus.Errorf("Update device %s status error", oldBd.Name)
-				s.Blockdevices.Enqueue(s.Namespace, newBd.Name)
-			}
+		}
+		logrus.Infof("Try to update the device path of %s, which is changed from %s to %s.", oldBd.Name, oldBd.Status.DeviceStatus.DevPath, newBd.Status.DeviceStatus.DevPath)
+		oldBd.Status.DeviceStatus.DevPath = newBd.Status.DeviceStatus.DevPath
+		if _, err := s.Blockdevices.Update(oldBd); err != nil {
+			logrus.Errorf("Update device %s status error, wake up scanner again: %v", oldBd.Name, err)
+			s.Cond.Signal()
```
tserong, line 141: Why is this changed from `s.Blockdevices.Enqueue()` to `s.Cond.Signal()`?

tserong (approved):
LGTM. Just one question...

pkg/controller/blockdevice/scanner.go
Vicente-Cheng: The previous one is wrong because the Enqueue (will trigger the OnChange) can not ensure updating the device path. Wake up the scanner again to ensure we scan all devices again for the device status update.

brandboat: Correct me if I'm wrong, but when duplicate `status.deviceStatus.devPath` occurs, it means there are multiple different BlockDevice CRs with the same `status.deviceStatus.devPath`, right? This PR aims to ensure that the CR update is triggered, making sure that the devPath of the old BlockDevice is correctly updated. Is my understanding correct?

Vicente-Cheng: yeah, this might confuse the user to add disk on the UI. For the controller behavior, we have the general function to retrieve the device path for format/mount  (REF: https://github.com/harvester/node-disk-manager/blob/master/pkg/provisioner/common.go#L213).

This PR ensures that the device status reflects the correct path. (For a better UI view)

brandboat (approved):
LGTM, thanks!

Vicente-Cheng:
@mergifyio backport v0.7.x

PR #201, opened 2025-06-25
Fix LVM provisioner panic when Spec.Devices map is nil

jjsair0412:
**Problem:**

The LVM provisioner crashes with a panic "assignment to entry in nil map" when attempting to add a new device to an existing LVMVolumeGroup that has a nil Spec.Devices map. This happens because:

1. An existing LVMVolumeGroup CRD exists in the cluster
2. The Spec.Devices map field was never initialized (remains nil)
3. When adding a new device, the code tries to assign to `lvmVGCpy.Spec.Devices[deviceName] = devicePath`
4. Go panics because you cannot assign to entries in a `nil` map

This is a critical production bug that crashes the node-disk-manager pod, preventing LVM volume management.

**Stack trace location**: pkg/provisioner/lvm.go:190 in addDevOrCreateLVMVgCRD function

**Solution:**

Add nil checks and initialization for the Spec.Devices map before attempting to assign values to it. The fix:

1. Checks if `lvmVG.Spec.Devices` is nil before checking for existing devices
2. Initializes the map with `make(map[string]string)` if it's `nil`
3. Adds an additional safety check after `DeepCopy()` since the copied object may also have a `nil` map
4. Only then proceeds with the safe assignment `lvmVGCpy.Spec.Devices[l.device.Name] = l.device.Status.DeviceStatus.DevPath`

This ensures backward compatibility with existing LVMVolumeGroup CRDs that may have uninitialized Devices maps while preventing the panic.

**Related Issue:**
This addresses a production bug encountered during LVM disk management operations. No existing GitHub issue was filed prior to this fix.

**Test plan:**

- [x]  Identified exact panic location in stack trace
- [x]  Verified the fix addresses the nil map assignment issue
- [x]  Code change is minimal and focused on the specific problem
- [x]  Maintains backward compatibility with existing LVMVolumeGroup CRDs
- [x]  Testing in development environment (manual testing of LVM device addition)
- [x]  CI tests pass

Manual test scenario:

1. Create an LVMVolumeGroup with uninitialized Spec.Devices
2. Add a new block device with LVM provisioner
3. Verify no panic occurs and device is successfully added to the volume group

### DIFF
--- a/pkg/provisioner/lvm.go
+++ b/pkg/provisioner/lvm.go
@@ -182,11 +182,17 @@ func (l *LVMProvisioner) addDevOrCreateLVMVgCRD(lvmVG *diskv1.LVMVolumeGroup, fo
 		err = fmt.Errorf("failed to get LVMVolumeGroup %s, but notFound is False", l.vgName)
 		return
 	}
+	if lvmVG.Spec.Devices == nil {
+		lvmVG.Spec.Devices = make(map[string]string)
+	}
 	if _, found := lvmVG.Spec.Devices[l.device.Name]; found {
 		logrus.Infof("Skip this round because the devices are not changed")
 		return
 	}
 	lvmVGCpy := lvmVG.DeepCopy()
+	if lvmVGCpy.Spec.Devices == nil {
+		lvmVGCpy.Spec.Devices = make(map[string]string)
+	}
 	lvmVGCpy.Spec.Devices[l.device.Name] = l.device.Status.DeviceStatus.DevPath
 	if !reflect.DeepEqual(lvmVG, lvmVGCpy) {
 		if _, err = l.vgClient.Update(lvmVGCpy); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/provisioner/lvm.go
```diff
@@ -182,11 +182,17 @@ func (l *LVMProvisioner) addDevOrCreateLVMVgCRD(lvmVG *diskv1.LVMVolumeGroup, fo
...
 	}
+	if lvmVG.Spec.Devices == nil {
+		lvmVG.Spec.Devices = make(map[string]string)
+	}
 	if _, found := lvmVG.Spec.Devices[l.device.Name]; found {
 		logrus.Infof("Skip this round because the devices are not changed")
 		return
 	}
 	lvmVGCpy := lvmVG.DeepCopy()
+	if lvmVGCpy.Spec.Devices == nil {
```
votdev, line 193: Is this really necessary here? `lvmVG.Spec.Devices` is set up some lines above and should be copied by [DeepCopy](https://github.com/harvester/node-disk-manager/blob/321fe1d8f7f0f864abd34d7c7b284776d301b083/pkg/apis/harvesterhci.io/v1beta1/zz_generated_deepcopy.go#L400) to `lvmVGCpy`.

Vicente-Cheng (approved):
nice catch, thanks!

brandboat (approved):
LGTM, I will merge this PR after CI finished.

PR #206, opened 2025-07-02
Node-Disk-Manager should be able to clean up the out-of-date LVM

votdev:
**Problem:**
Adding a disk to a host/node fails when the disk is wiped with wipefs when the disk was previously used in another LVM setup.

**Solution:**
Cleanup existing LVM artefacts in a pre-format step.

**Related Issue:**
https://github.com/harvester/harvester/issues/7961

**Test plan:**
ToDo...


pkg/provisioner/longhornv2.go
```diff
@@ -61,8 +60,9 @@ func NewLHV2Provisioner(
 // nvme bdev driver, device activation will fail if there's an existing
 // filesystem on the device, so we need to make sure to wipe before use.
 func (p *LonghornV2Provisioner) Format(devPath string) (isFormatComplete, isRequeueNeeded bool, err error) {
-	if _, err = utils.NewExecutor().Execute("wipefs", []string{"-a", devPath}); err != nil {
-		return false, false, err
+	err = p.wipeDevice(devPath)
+	if err != nil {
+		return false, true, err
```
tserong, line 65: This is returning isRequeueNeeded == true if wipefs fails, which I assume means this will be called repeatedly until wipefs succeeds? The old behaviour was to return isRequeueNeeded == false, so I just wanted to double check this.

...

Oh, wait, I've just discovered the LHv1 Provisioner Format function already returns isRequeueNeeded == true if wipefs fails, so I guess this change is just fixing the discrepancy between the two Format functions?

PR #209, opened 2025-07-08
fix: skip duplicate disk check if device has no WWN

tserong:
**Problem:**
In cases where disks are manually formatted to work around having no WWN, they get caught by the duplicate WWN checker.  This prevents adding multiple disks of this kind.

**Solution:**
In this case, disks are identified by FS UUID, which means we can safely skip the duplicate check anyway.

**Related Issue:**
https://github.com/harvester/harvester/issues/7173

**Test plan:**
- Deploy a virtualized harvester cluster. One node is fine.
- Add two extra virtio disks (/dev/vdb, /dev/vdc) - these will not have WWNs.
- Initially, these two disks should not be visible in the Harvester UI if you try to add extra disks to a node.
- Run `mkfs /dev/vdb ; mkfs /dev/vdc` to make filesystems on those disks, then `kubectl -n harvester-system rollout restart daemonset/harvester-node-disk-manager` to make NDM rescan the disks
- If this fix is not present, running `kubectl -n longhorn-system get bds` should only show one of `/dev/vdb` or `/dev/vdc`. The NDM logs (`kubectl -n harvester-system logs -l app.kubernetes.io/name=harvester-node-disk-manager --tail=-1`) will include a warning for the other device, like: `Skip adding duplicated WWN device unknown: /dev/vdc`. Likewise the Harvester UI will only show one of these disks available to add.
- With this fix in place, both `/dev/vdb` and `/dev/vdc` should appear in the list of BDs, and both should be available to add in the Harvester GUI. The NDM logs should say `Skipping duplicate disk check (device has no WWN)"` instead of the warning mentioned above.

### DIFF
--- a/pkg/controller/blockdevice/scanner.go
+++ b/pkg/controller/blockdevice/scanner.go
@@ -199,17 +199,25 @@ func (s *Scanner) scanBlockDevicesOnNode() error {
 			// remove blockdevice from old device so we can delete missing devices afterward
 			delete(oldBds, bd.Name)
 		} else {
-			/*
-			 * Prevent add duplicated wwn even if the device path is different.
-			 * That prevent the device from being formatted again.
-			 * We should use multiple device path (TBD) for the same wwn.
-			 */
-			if slices.Contains(existingWWNs, bd.Status.DeviceStatus.Details.WWN) {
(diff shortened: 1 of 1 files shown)


Vicente-Cheng (approved):
lgtm, thanks!

PR #211, opened 2025-07-10
Block auto provision during upgrade

Vicente-Cheng:
**Problem:**
Auto-provisioning the disk would bring the potential risk during the upgrade.
Like disk data erased.

**Solution:**
block the auto-provision during upgrade

**Related Issue:**
https://github.com/harvester/harvester/issues/8296

**Test plan:**
1. create one node cluster
2. add extra disk
3. create upgrade
4. add the auto-provision path for the extra disk
5. the disk should not be provisioned


Vicente-Cheng:
please ignore the codeFactor complaint, just related to the comment format

tserong (approved):
LGTM

pkg/controller/blockdevice/controller.go
```diff
@@ -27,6 +30,7 @@ import (
 
 const (
 	blockDeviceHandlerName = "harvester-block-device-handler"
+	harvesterSystemNS      = "harvester-system"
```
brandboat, line 33: nit: what about using `utils.HarvesterNS`?

pkg/controller/blockdevice/controller.go
```diff
@@ -429,6 +437,47 @@ func (c *Controller) updateAutoProvisionDevice(device *diskv1.BlockDevice) (*dis
...
+		return true
+	}
+
+	return !isOnUpgrade(upgradesList)
+}
+
+func isOnUpgrade(upgrades []*harvesterv1.Upgrade) bool {
+	for _, upgrade := range upgrades {
+		cond := getUpgradeCondition(upgrade.Status.Conditions, harvesterv1.UpgradeCompleted)
+		if cond != nil && cond.Status != corev1.ConditionTrue {
```
brandboat, line 461: If upgrade failed, we won't do disk auto provision, right?
I'm just thinking a scenario that if there are some upgrade CRs existing, and there is one legacy upgrade CR that stays in completed failed state, then auto disk provision won't trigger.

Vicente-Cheng: yeah, but iiuc, we wouldn't keep the failed upgrade CR.
We need to delete the failed upgrade CR, and then we can trigger another one.

cc @starbops, would the above situation have happened?

brandboat: > yeah, but iiuc, we wouldn't keep the failed upgrade CR.
We need to delete the failed upgrade CR, and then we can trigger another one.

I just tried, and we can perform an upgrade after a failed upgrade.

BTW, we can leverage on the `harvesterhci.io/upgradeState` to check if there is any ongoing upgrade. We already have similar logic in 
https://github.com/harvester/harvester/blob/c1232fe90913b9755f154f23139314f98f25c1a2/pkg/webhook/resources/upgrade/validator.go#L129-L141

pkg/controller/blockdevice/controller.go
Vicente-Cheng: updated

tserong (approved):

brandboat (approved):
LGTM, thanks!

PR #219, opened 2025-08-01
controller: use upgrade client instead controller cache

Vicente-Cheng:
- We did not watch Factory, the cache would not be synced. Use the client directly.

**Problem:**
The upgrade cache did not sync forever

**Solution:**
Use the client directly

**Related Issue:**
https://github.com/harvester/harvester/issues/8296

**Test plan:**


Yu-Jack (approved):
lgtm

PR #221, opened 2025-08-19
feat: support identifying multipath device

Yu-Jack:
**Problem:**
We don't identify /dev/dm-x (multipath device).

**Solution:**
Using `multipath -C` and `multipath -c` to identify whether the block device is multipath device or not.

**Related Issue:**
https://github.com/harvester/harvester/issues/6975

**Test plan:**

1. Create a disk.
```
sudo qemu-img create -f raw /var/lib/libvirt/images/mydisk1.img 5G
```

2. Pass it twice with same serial ID when installing.
```
sudo virt-install \ 
--name test5 \
--memory {memory} \
--vcpus {cpu} \
--os-variant=slem5.3 \
--disk /var/lib/libvirt/images/test5.qcow2,size=300,bus=virtio,format=qcow2 \

# You must use it twice to simulate two disk paths.
--disk path=/var/lib/libvirt/images/mydisk1.img,bus=scsi,cache=none,shareable=on,serial=disk1 \
--disk path=/var/lib/libvirt/images/mydisk1.img,bus=scsi,cache=none,shareable=on,serial=disk1 \

--network network=default \
--graphics vnc,listen=0.0.0.0,password=1234 \
--iommu model=intel,driver.intremap=on,driver.caching_mode=on \
--features ioapic.driver=qemu \
--pxe --check disk_size=off
```

3. Enable and start multipathd with following config.

```
# 3.1 /etc/multipath.conf
blacklist {
    wwid ".*"
}
blacklist_exceptions {
    wwid "^0QEMU_QEMU_HARDDISK_disk[0-9]+"
}

# 3.2 command
systemctl enable multipathd
systemctl start multipathd

# 3.3 Use following command to persist the multipath.conf in /oem/90_custom.yaml
awk 'BEGIN{mp=0;cs=0} /- path: \/etc\/multipath\.conf/{mp=1;print;next} mp&&/content: \|-/{cs=1;print "              content: |-";print "                blacklist {";print "                    device {";print "                        vendor \"!QEMU\"";print "                        product \"!QEMU HARDDISK\"";print "                    }";print "                    wwid \".*\"";print "                }";print "                blacklist_exceptions {";print "                    wwid \"^0QEMU_QEMU_HARDDISK_disk[0-9]+\"";print "                }";next} mp&&cs&&/^              [^ ]/{mp=0;cs=0;print;next} mp&&cs{next} {print}' /oem/90_custom.yaml > /oem/90_custom.yaml.tmp && mv /oem/90_custom.yaml.tmp /oem/90_custom.yaml
```

4. You should see there are available multipath devices. Add it. Also, please add another different disk with Longhorn V2. because Longhorn V2 also uses `/dev/dm-`. So, we need to test this case.

    image

    image

6. Start a virtual machine and try to schedule the replica on this new storage. (Try to enable `Soft Anti Affinity` in longhorn side)

6. Reboot the node. Everything should be fine.

### DIFF
--- a/pkg/controller/blockdevice/scanner.go
+++ b/pkg/controller/blockdevice/scanner.go
@@ -156,13 +156,20 @@ func (s *Scanner) handleExistingDev(oldBd *diskv1.BlockDevice, newBd *diskv1.Blo
 		// If system doesn't enable multipathd, the block device will become inactive after reboot.
 		// So, we need to add a checking after starting multipathd.
 		// Then the block device will become active from inactive.
+		oldBdCp := oldBd.DeepCopy()
+
 		if _, err := utils.IsMultipathDevice(oldBd.Status.DeviceStatus.DevPath); err == nil {
 			logrus.Infof("The multipath device %s is available, change it to active.", oldBd.Name)
-			oldBd.Status.State = diskv1.BlockDeviceActive
+			oldBdCp.Status.State = diskv1.BlockDeviceActive
 		}
 
-		if _, err := s.Blockdevices.Update(oldBd); err != nil {
-			logrus.Errorf("Update device %s status error, wake up scanner again: %v", oldBd.Name, err)
+		if reflect.DeepEqual(oldBd, oldBdCp) {
+			logrus.Debugf("Skip updating multipath device %s", oldBd.Name)
+			return
+		}
+
+		if _, err := s.Blockdevices.Update(oldBdCp); err != nil {
+			logrus.Errorf("Update device %s status error, wake up scanner again: %v", oldBdCp.Name, err)
 			s.Cond.Signal()
 		}
 	} else {


pkg/utils/utils.go
```diff
@@ -220,3 +220,39 @@ func CallerWithCondLock[T any](cond *sync.Cond, f func() T) T {
...
+
+	// Execute multipath -c /dev/xxx command to check if device is managed by multipath
+	// multipath -c returns exit code 0 if device belongs to multipath, non-zero if not
+	output, err := executor.Execute("multipath", []string{"-c", fmt.Sprintf("/dev/%s", deviceName)})
+	if err != nil {
+		return output, err
+	}
+
+	return output, nil
+}
```
Yu-Jack, line 258: Although we can check `/sys/block/xx/holders` or `/sys/block/xx/slaves`, I don't recommend that we handcraft searching ourselves if there is already a wrapped command unless it's actually costly.

ibrokethecloud (approved):
lgtm. thanks.

tserong (approved):
LGTM - thanks!

PR #244, opened 2025-09-19
Validate single node disk removals

martindekov:
Rejecting disk removals when there are attached volumes in Degraded
state on single node harvester set ups.

We also make sure to reject removals when block devices are
in failed state as well.

The logic is split in two parts:
* We take the disk selectors of volumes in degraded state and
compare the selectors against the disktags
* We take the longhorn node related with the current blockdevice 
and through the node we pick the backing image all through indexing

In both scenarios in case there is match while disk is being
disabled, we list to the user which volumes and then images
needs to get fixed.

Signed-off-by: Martin Dekov <martin.dekov@suse.com>

**Problem:**
Users can directly delete disks without any warnings explained in issue:
https://github.com/harvester/harvester/issues/3344

**Solution:**
Reject deletions of disks when there are degraded volumes when there is only single node harvester cluster

**Related Issue:**

https://github.com/harvester/harvester/issues/3344

**Test plan:**

Unit testing added

Replicate Degraded Volume Scenario:
1. Create single node harvester cluster
2. Attach SATA disks before installation
3. Once installed add the disks through the Hosts -> Edit Config -> Storage -> Add Disk dropdown
4. Tag the added disk with `new` tag for example
5. Create storage class with 3 replicas and attach the disk Advanced -> Storage Classes -> Create -> Disk Selector -> `new` (same as disk tag above)
6. Create virtual machine and attach volume to use that storage class
7. Since replicas are 3 while we have only 1 volume we end up with Degraded volume
8. Try to delete the disk - see rejection

Replicate Failed Backing Image scenario:
1. Follow steps 1 to 4 above
2. In images tab click Create button to create a new image
3. While creating pick file image and click Storage Class tab and pick storage class created from above - "new"
4. While image is uploaded refresh the page multiple times to provoke EOF upload error
5. Try deleting the disk from Hosts page - check error
> Note: make sure the virtualmachineimage has block device as backend before testing so when image is created corresponding blockdevice is created as well

### DIFF
--- a/pkg/webhook/blockdevice/validator.go
+++ b/pkg/webhook/blockdevice/validator.go
@@ -277,8 +277,8 @@ func backingImageByDiskUUIDIndexer(bi *lhv1.BackingImage) ([]string, error) {
 	if bi.Spec.DiskFileSpecMap == nil {
 		return []string{}, nil
 	}
-	diskUUIDs := make([]string, 0, len(bi.Spec.DiskFileSpecMap))
-	for key := range bi.Spec.DiskFileSpecMap {
+	diskUUIDs := make([]string, 0, len(bi.Status.DiskFileStatusMap))
+	for key := range bi.Status.DiskFileStatusMap {
 		diskUUIDs = append(diskUUIDs, key)
 	}
 	return diskUUIDs, nil


pkg/webhook/blockdevice/validator.go
```diff
@@ -62,6 +73,37 @@ func (v *Validator) validateProvisioner(bd *diskv1.BlockDevice) error {
 	return nil
 }
 
+func (v *Validator) validateDisk(old, new *diskv1.BlockDevice) error {
+	if old.Spec.Provision && !new.Spec.Provision {
+		nodeList, err := v.nodeCache.List(labels.Everything())
+		if err != nil {
+			return err
+		}
+		if len(nodeList) > 1 {
```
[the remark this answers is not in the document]
martindekov: Fixed so we add logic in case of a single node.

pkg/webhook/blockdevice/validator.go
```diff
@@ -62,6 +73,37 @@ func (v *Validator) validateProvisioner(bd *diskv1.BlockDevice) error {
...
+				return err
+			}
+			for _, pvc := range pvcList {
+				for _, diskTag := range old.Spec.Tags {
+					if diskTag == pvc.Spec.StorageClassName {
+						volume, err := v.volumeCache.Get("longhorn-system", pvc.Name)
+						if err != nil {
+							return err
+						}
+						if volume.Status.Robustness == lhv1.VolumeRobustnessDegraded {
```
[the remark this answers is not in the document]
martindekov: Refactored so no more nested for loops, but rather sequence of filtration on objects as per your suggestion

pkg/webhook/blockdevice/validator.go
```diff
@@ -62,6 +73,37 @@ func (v *Validator) validateProvisioner(bd *diskv1.BlockDevice) error {
 	return nil
 }
 
+func (v *Validator) validateDisk(old, new *diskv1.BlockDevice) error {
+	if old.Spec.Provision && !new.Spec.Provision {
```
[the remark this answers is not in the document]
martindekov: Added nil check and in case LonghornProvisioner is not set we skip validation

pkg/webhook/blockdevice/validator.go
```diff
@@ -127,6 +165,70 @@ func (v *Validator) validateVGIsAlreadyUsed(bd *diskv1.BlockDevice) error {
...
+	}
+	selectorDegradedVol := make(map[string]string)
+	for _, pv := range pvList {
+		if degradedVolumes[pv.Name] {
+			diskSelector := pv.Spec.CSI.VolumeAttributes[DiskSelectorKey]
+			if len(diskSelector) != 0 {
+				selectorDegradedVol[pv.Spec.CSI.VolumeAttributes[DiskSelectorKey]] = pv.Name
+			}
+		}
+	}
```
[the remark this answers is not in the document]
martindekov: Thanks Moritz this:
> Given that the .Get() is an operation on a local cache

Was the reason I listed and not use `Get` as I was not sure whether the cache is local or remote, e.g. if we execute api call against kubernetes to load cache through this interface.

If this is against local cache and all api calls would be internal, if any, yes I will switch to that, thanks for the suggestion and clarification. Also I will take general note on this, was not aware.

pkg/webhook/blockdevice/validator.go
```diff
@@ -127,6 +167,70 @@ func (v *Validator) validateVGIsAlreadyUsed(bd *diskv1.BlockDevice) error {
...
+
+func (v *Validator) validateFailedVMImages(old *diskv1.BlockDevice) error {
+	vmImageList, err := v.vmImageCache.List("", labels.Everything())
+	if err != nil {
+		return err
+	}
+	selectorFailedVMIMage := make(map[string]string)
+	for _, vmImage := range vmImageList {
+		if vmImage.Status.Failed != 0 && len(vmImage.Spec.StorageClassParameters[DiskSelectorKey]) > 0 {
+			selectorFailedVMIMage[vmImage.Spec.StorageClassParameters[DiskSelectorKey]] = vmImage.Spec.DisplayName
```
w13915984028, line 218: define `vmImage.Spec.StorageClassParameters[DiskSelectorKey]` as a local var, it occurs twice and it is a map looking

pkg/webhook/blockdevice/validator.go
```diff
@@ -127,6 +167,70 @@ func (v *Validator) validateVGIsAlreadyUsed(bd *diskv1.BlockDevice) error {
...
+		}
+	}
+	failedVMImages := ""
+	for _, diskTag := range old.Spec.Tags {
```
w13915984028, line 222: if the `old.Spec.Tags` can be empty slice/map, it coud be checked at very first to avoid above searching

w13915984028 (commented):
some suggestions to optimize the code

pkg/webhook/blockdevice/validator.go
```diff
@@ -127,6 +167,70 @@ func (v *Validator) validateVGIsAlreadyUsed(bd *diskv1.BlockDevice) error {
 	return nil
 }
 
+func (v *Validator) validateDegradedVolumes(old *diskv1.BlockDevice) error {
+	volumeList, err := v.volumeCache.List("longhorn-system", labels.Everything())
+	if err != nil {
+		return err
+	}
+	degradedVolumes := make(map[string]bool)
```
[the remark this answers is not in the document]
martindekov: switched to map[string]struct{} :heavy_check_mark:

pkg/webhook/blockdevice/validator.go
```diff
@@ -127,6 +167,70 @@ func (v *Validator) validateVGIsAlreadyUsed(bd *diskv1.BlockDevice) error {
...
+		if vol.Status.Robustness == lhv1.VolumeRobustnessDegraded {
+			degradedVolumes[vol.Name] = true
+		}
+	}
+	if len(degradedVolumes) == 0 {
+		return nil
+	}
+	selectorDegradedVol := make(map[string]string)
+	for name, degraded := range degradedVolumes {
+		if degraded {
```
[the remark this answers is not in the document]
martindekov: simplified
```
	for name, degraded := range degradedVolumes {
		if degraded {
```
to:
```
	for name, _ := range degradedVolumes {
```

as per above suggestions :heavy_check_mark:

pkg/webhook/blockdevice/validator.go
```diff
@@ -127,6 +167,70 @@ func (v *Validator) validateVGIsAlreadyUsed(bd *diskv1.BlockDevice) error {
...
+	}
+	return nil
+}
+
+func (v *Validator) validateFailedVMImages(old *diskv1.BlockDevice) error {
+	vmImageList, err := v.vmImageCache.List("", labels.Everything())
+	if err != nil {
+		return err
+	}
+	selectorFailedVMIMage := make(map[string]string)
```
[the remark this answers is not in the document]
martindekov: good suggestion, added :heavy_check_mark:

pkg/webhook/blockdevice/validator.go
```diff
@@ -62,6 +76,28 @@ func (v *Validator) validateProvisioner(bd *diskv1.BlockDevice) error {
...
 
+func (v *Validator) validateLHDisk(oldBd, newBd *diskv1.BlockDevice) error {
+	if oldBd.Spec.Provisioner != nil && newBd.Spec.Provisioner != nil &&
+		oldBd.Spec.Provisioner.Longhorn != nil && newBd.Spec.Provisioner.Longhorn != nil &&
+		oldBd.Spec.Provision && !newBd.Spec.Provision {
+		nodeList, err := v.nodeCache.List(labels.Everything())
+		if err != nil {
+			return err
+		}
+		if len(nodeList) == 1 && len(oldBd.Spec.Tags) > 0 {
```
w13915984028, line 87: even when len(nodeList) > 1, the check might still be necessary, (1) the `witness` node doesn't run service
(2) if SC other volume has node related selector

just check disk removal, skip the node count evaluation

pkg/webhook/blockdevice/validator.go
```diff
@@ -127,6 +163,78 @@ func (v *Validator) validateVGIsAlreadyUsed(bd *diskv1.BlockDevice) error {
...
+		pv, err := v.pvCache.Get(name)
+		if err != nil {
+			return err
+		}
+		diskSelector := ""
+		if pv.Spec.CSI != nil {
+			diskSelector = pv.Spec.CSI.VolumeAttributes[utils.DiskSelectorKey]
+		}
+		if len(diskSelector) != 0 {
+			selectorDegradedVol[diskSelector] = pv.Name
```
w13915984028, line 194: if multi PVs have the same `diskSelector`, will here only the last pv.Name is recorded?

pkg/webhook/blockdevice/validator.go
```diff
@@ -127,6 +163,78 @@ func (v *Validator) validateVGIsAlreadyUsed(bd *diskv1.BlockDevice) error {
...
+		}
+	}
+	degradedVolString := ""
+	for _, diskTag := range old.Spec.Tags {
```
w13915984028, line 198: seems you need to get/use `old.Spec.Tags` map on above L194 to check and format the string directly

pkg/webhook/blockdevice/validator.go
```diff
@@ -127,6 +163,78 @@ func (v *Validator) validateVGIsAlreadyUsed(bd *diskv1.BlockDevice) error {
...
+	if err != nil {
+		return err
+	}
+	if len(vmImageList) == 0 {
+		return nil
+	}
+	selectorFailedVMIMage := make(map[string]string)
+	for _, vmImage := range vmImageList {
+		diskSelectorValue := vmImage.Spec.StorageClassParameters[utils.DiskSelectorKey]
+		if vmImage.Status.Failed != 0 && len(diskSelectorValue) > 0 {
```
w13915984028, line 221: put `if vmImage.Status.Failed == 0 { continue }` before L220

w13915984028 (commented):
some suggestions

pkg/webhook/blockdevice/validator.go
```diff
@@ -127,6 +163,78 @@ func (v *Validator) validateVGIsAlreadyUsed(bd *diskv1.BlockDevice) error {
...
+			selectorFailedVMIMage[diskSelectorValue] = vmImage.Spec.DisplayName
```
w13915984028, line 222: similar, if multi images have same disk selector, only the last one will be booked on the map ?

w13915984028 (commented):
comments

pkg/webhook/blockdevice/validator.go
martindekov: As per offline discussion, yes, appending names to the key as opposed to using the last until user fixes all, now user knows all problematic :heavy_check_mark:

pkg/webhook/blockdevice/validator.go
martindekov: We agreed that 1 node setup is problematic, for other set ups we won't reject forcefully through webhook, but rather warn in UI. If I follow with this suggestion the UI warning change becomes obsolete so I think we need to go back to discussions if we want to revert to that.

Will leave the thread open if you insist on resurrecting the discussion

pkg/webhook/blockdevice/validator.go
martindekov: If you suggest I can build the string in the loop above, I did that to avoid nested loops (tags are not map, but slice), though originally I did that, will leave discussion open in case you think merging the loops would be the better approach

pkg/webhook/blockdevice/validator.go
martindekov: Comment above was not relevant, now in the value we contain all names by appending to the key :heavy_check_mark:

pkg/webhook/blockdevice/validator.go
w13915984028: then keep it for now; if mulit nodes have similare issue, fix via other PRs

pkg/webhook/blockdevice/validator.go
w13915984028: the new implementation is also ok, which stores all names under same key

pkg/webhook/blockdevice/validator.go
```diff
@@ -127,6 +163,81 @@ func (v *Validator) validateVGIsAlreadyUsed(bd *diskv1.BlockDevice) error {
...
 
+func (v *Validator) validateDegradedVolumes(old *diskv1.BlockDevice) error {
+	volumeList, err := v.volumeCache.List(utils.LonghornSystemNamespaceName, labels.Everything())
+	if err != nil {
+		return err
+	}
+	if len(volumeList) == 0 {
+		return nil
+	}
+	degradedVolumes := make(map[string]struct{})
```
m-ildefons, line 174: I'm not sure why this is a map. To me it reads as if this could be a slice, e.g.

```golang
degradedVolumes := make([]string{}, 32)
for _, vol := range volumeList {
  if vol.Status.Robustness == lhv1.VolumeRobustnessDegraded {
    degradedVolumes = append(degradedVolumes, vol.Name)
  }
}
[...]
for i := range degradedVolumes {
  pv, err := v.pvCache.Get(degradedVolumes[i])
  [...]
}
```
Is there any advantages to using a map instead of a slice?
Afaik maps are hashmaps, which have constant time complexity operations. However the operations are generally slower than on an array - as long as the array doesn't need to be resized. Conversely iterating over a slice is usually faster.
Therefore I wonder if we shouldn't use a reasonably-sized slice here. I used 32 in my example, since that should allow for `degradedVolumes` to remain on the stack.

deploy/charts/harvester-node-disk-manager/templates/rbac.yaml
```diff
@@ -68,6 +68,9 @@ rules:
   - apiGroups: [ "admissionregistration.k8s.io" ]
     resources: [ "validatingwebhookconfigurations", "mutatingwebhookconfigurations" ]
     verbs: [ "*" ]
+  - apiGroups: [ "" ]
+    resources: [ "nodes" ]
+    verbs: [ "get", "watch", "list" ]
```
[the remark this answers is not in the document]
martindekov: Added :heavy_check_mark:

scripts/unit-test
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+set -e
+
+cd $(dirname $0)/..
+
+echo Running unit tests
+go test -v -cover ./pkg/...
```
m-ildefons, line 7: Just a nit, you don't have to correct this:

If you want coverage output from the containerized build (i.e. using `make unit-test`), then you need to do some additional things:

1) specify a coverage output file
```bash
go test -v -cover -coverprofile=coverage.out ./pkg/...
```

2) do the conversion to the final output format (e.g. HTML) inside the container
```bash
go test [...]
go tool cover -html=coverage.out -o coverage.html
```

3) add the final coverage output file in `Dockerfile.dapper` where the environment variable `DAPPER_OUTPUT` is defined, e.g.:
```Dockerfile
[...]
ENV DAPPER_OUTPUT ./bin ./manifests ./pkg ./coverage.html
```

Then you can do `make unit-test`, the coverage report will generated and placed outside of the build container so you can analyze it.

m-ildefons (approved):
LGTM.

scripts/unit-test
martindekov: Thanks Moritz, I will leave this open. I intentionally made it simple, didn't want to overcrowd this change as it already contains too much different things. Augmenting pipeline to output coverage in the right way is a work on it's own. Also I think there were badges to show coverage on front page etc etc. 

We can open an separate issue for this I think in this repo? WDYT?

pkg/webhook/blockdevice/validator.go
```diff
@@ -127,6 +163,81 @@ func (v *Validator) validateVGIsAlreadyUsed(bd *diskv1.BlockDevice) error {
...
+			failedVMImages += fmt.Sprintf(" %s: %v", diskTag, val)
+		}
+	}
+	if len(failedVMImages) > 0 {
+		return fmt.Errorf("the following tags referenced by virtualmachineimages: %s attached to disk: %s are in failed state; evict disk before proceeding",
+			failedVMImages, old.Spec.DevPath)
+	}
+	return nil
+}
+
```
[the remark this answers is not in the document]
WebberHuang1118: IMO, you can use the mount point described in the Longhorn node to correlate it with the BlockDevice CR. You might also consider using a cache indexer to identify the backing images that need to be checked, rather than tediously iterating through the entire data structure.

WebberHuang1118 (changes requested):
Some of the logic for checking backing images seems incorrect to me; please see my comment. 

Also suggest reviewing the [Backing Image LEP](https://github.com/longhorn/longhorn/blob/master/enhancements/20210701-backing-image.md) for the relevant data structures.

Thanks

pkg/webhook/blockdevice/validator.go
```diff
@@ -127,6 +173,81 @@ func (v *Validator) validateVGIsAlreadyUsed(bd *diskv1.BlockDevice) error {
...
+		if val, ok := selectorDegradedVol[diskTag]; ok {
+			degradedVolString += fmt.Sprintf(" %s: %v", diskTag, val)
+		}
+	}
+	if len(degradedVolString) > 0 {
+		return fmt.Errorf("the following tags with volumes:%s attached to disk: %s are in degraded state; evict disk before proceeding",
+			degradedVolString, old.Spec.DevPath)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
martindekov: selector is how we trace which volumes are attached to the disk we currently we are trying to remove so:
1. Tagging a disk
2. We use that tag to create storage class 

And this tag is used as selector across those components. That way If I Delete the disk I trace back the volumes created through the `disk -> tag -> sc` way

WebberHuang1118:
Hi @martindekov 

The original issue [#3344](https://github.com/harvester/harvester/issues/3344) should also cover multi-node environments.

As I see it, this PR is attempting to address [#3344](https://github.com/harvester/harvester/issues/3344), so the PR shouldn’t be limited to single-node setups. From my perspective, [#3344](https://github.com/harvester/harvester/issues/3344) and [#9218](https://github.com/harvester/harvester/issues/9218) appear to be duplicates.

IMO, we should design a unified solution that applies to both single-node and multi-node environments in this PR. There’s no need to separate it into another issue ([#9218](https://github.com/harvester/harvester/issues/9218)) and another PR.

martindekov:
To re-iterate, generally we agreed the solution for all node set ups is warning, not webhook rejection. Webhook rejection is for the corner case where we cannot really have replicated storage even if we set the affinity rule as per the comment:
https://github.com/harvester/harvester/issues/3344#issuecomment-3301636346

The issue - https://github.com/harvester/harvester/issues/9218 - refers to the Warning/UI
The issue - https://github.com/harvester/harvester/issues/3344 - was refering to both, but reduced the scope as this PR with the webhook took too much time which I did not anticipate.

So the work between the issues is split - [one for the UI](https://github.com/harvester/harvester/issues/9218) and one for [Backend](https://github.com/harvester/harvester/issues/3344).

If we decide all warning work to go to the webhook backend, and leave the front end warning aspect, we also need to find ways to force delete disks in bad state as it will be rejected on API level. This design would further extend the scope of the issue.

WebberHuang1118:
> To re-iterate, generally we agreed the solution for all node set ups is warning, not webhook rejection. Webhook rejection is for the corner case where we cannot really have replicated storage even if we set the affinity rule as per the comment:
> https://github.com/harvester/harvester/issues/3344#issuecomment-3301636346

Not every detail becomes clear until the PR is created. IMO, for reviewers, changes or updates to the original design or proposal are sometimes necessary.


> The issue - https://github.com/harvester/harvester/issues/9218 - refers to the Warning/UI
> The issue - https://github.com/harvester/harvester/issues/3344 - was refering to both, but now refers only to webhook/backend work - this PR as it took too much time which I did not anticipate.
> 
> So the work between the issues is split - https://github.com/harvester/harvester/issues/9218 and one for https://github.com/harvester/harvester/issues/3344.

I agree that [#3344](https://github.com/harvester/harvester/issues/3344) targets the backend part only. However, based on my code reading, the current implementation only includes a webhook to reject disk removal for preventing data loss in a single-node cluster. This is not sufficient for [#3344](https://github.com/harvester/harvester/issues/3344) for the following reasons:

* **Multi-node clusters:** A volume may have only one healthy replica, and a backing image may have only one valid copy. Therefore, even in a multi-node cluster, a strict webhook block is still required to prevent data loss.
* **Single-node clusters:** Multiple disks can be attached to a single node, so separating the logic between single-node and multi-node environments isn’t necessary and could make the code path more complex. IMO, we should have a unified mechanism that works for all cases.
* **PR design scope:** If the PR sticks with the current design ([comment](https://github.com/harvester/harvester/issues/3344#issuecomment-3301636346)), it won’t fully resolve the root issue [#3344](https://github.com/harvester/harvester/issues/3344). Even if merged, the source issue likely won’t be marked as `ready-for-testing`, which may cause confusion for the QA team during verification.

**TL;DR**
I agree that [#9218](https://github.com/harvester/harvester/issues/9218) should focus on GUI warning messages — specifically, when disk removal will *not* lead to data loss (i.e., all volumes still have healthy replicas and all backing images have ready copies even if the target disk is removed).

Meanwhile, [#3344](https://github.com/harvester/harvester/issues/3344) should focus on the original context — if removing a disk would cause any volume to lose its only healthy replica or any backing image to lose the only ready copy, the webhook should strongly block the operation to prevent data loss, regardless of how many nodes are in the cluster.

If we continue with the current PR design, I believe the context and scope of [#3344](https://github.com/harvester/harvester/issues/3344) should at least be reduced — for example, to `[ENHANCEMENT] Delete disk operation should consider data availability for single-node Harvester`. However, to be honest, that doesn’t make much sense from my perspective.

Ping @Vicente-Cheng @ihcsim for your thoughts on this topic.

PR #317, opened 2026-06-29
feat: implement mount watcher for Linux and integrate with Udev monitor

Yu-Jack:
**Problem:**

When users manually mount a disk, the disk is still an available option in harvester host configuration.
Users might accidentally select the mounted disk as harvester backend storage.

**Solution:**

We should skip mounted disks.

**Related Issue:**

https://github.com/harvester/harvester/issues/8990

**Test plan:**

https://github.com/user-attachments/assets/3722e640-c899-4c05-ac95-6a6fc77c7971

### DIFF
--- a/pkg/udev/mount_watcher.go
+++ b/pkg/udev/mount_watcher.go
@@ -1,0 +1,76 @@
+//go:build linux
+
+package udev
+
+import (
+	"context"
+	"math"
+
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+
+	"github.com/harvester/node-disk-manager/pkg/utils"
+)
+
+// NDM mounts the host's /proc at /host/proc (see daemonset.yaml), so we watch
+// /host/proc/1/mountinfo (host PID 1 = init) to observe the host mount namespace.
+// Using /proc/self/mountinfo would only reflect the container's own namespace.
+const procMountInfo = "/host/proc/1/mountinfo"
+
+// watchMounts monitors mount table changes by polling /proc/self/mountinfo for POLLERR.
+// The kernel raises POLLERR on this fd whenever any mount or umount occurs.
+func (u *Udev) watchMounts(ctx context.Context) {
+	logrus.Debugf("mount watcher: opening %s", procMountInfo)
+	fd, err := unix.Open(procMountInfo, unix.O_RDONLY|unix.O_CLOEXEC, 0)
+	if err != nil {
+		logrus.Errorf("mount watcher: failed to open %s: %v", procMountInfo, err)
+		return
+	}
+	logrus.Debugf("mount watcher: opened fd=%d", fd)
+	defer func() {
+		logrus.Debugf("mount watcher: closing fd=%d", fd)
+		unix.Close(fd)
+	}()
+
+	if fd > math.MaxInt32 {
+		logrus.Errorf("mount watcher: fd %d out of int32 range", fd)
+		return
+	}
+	fdInt32 := int32(fd) //nolint:gosec // fd is guaranteed non-negative by Open and bounded by math.MaxInt32 check above
+
+	// POLLERR fires when the kernel marks /proc/self/mountinfo dirty (mount/umount)
+	fds := []unix.PollFd{{Fd: fdInt32, Events: unix.POLLERR}}
+
+	logrus.Infof("mount watcher: watching %s for mount table changes", procMountInfo)
+
+	for {
+		select {
+		case <-ctx.Done():
+			logrus.Debug("mount watcher: context cancelled, exiting")
+			return
+		default:
+		}
+
+		logrus.Debugf("mount watcher: polling fd=%d for POLLERR, timeout=5s", fd)
+		n, err := unix.Poll(fds, 5*1000)
+		if err != nil {
+			if err == unix.EINTR {
+				logrus.Debug("mount watcher: poll interrupted by signal, retrying")
+				continue
+			}
+			logrus.Errorf("mount watcher: poll error: %v", err)
+			return
+		}
+		if n == 0 {
+			logrus.Debug("mount watcher: poll timeout, no mount changes")
+			continue
+		}
+
+		logrus.Debugf("mount watcher: POLLERR received (revents=0x%x), mount table changed", fds[0].Revents)
+		utils.CallerWithCondLock(u.scanner.Cond, func() any {
+			logrus.Info("mount watcher: mount change detected, waking scanner")
+			u.scanner.Cond.Signal()
+			return nil
+		})
+	}
+}
--- a/pkg/udev/mount_watcher.go
+++ b/pkg/udev/mount_watcher.go
@@ -4,6 +4,7 @@ package udev
 
 import (
 	"context"
+	"fmt"
 	"math"
 
 	"github.com/sirupsen/logrus"
@@ -19,11 +20,12 @@ const procMountInfo = "/host/proc/1/mountinfo"
 
 // watchMounts monitors mount table changes by polling /proc/self/mountinfo for POLLERR.
 // The kernel raises POLLERR on this fd whenever any mount or umount occurs.
-func (u *Udev) watchMounts(ctx context.Context) {
+// On any unrecoverable error it sends to errChan so spawnMountWatcher can respawn it.
+func (u *Udev) watchMounts(ctx context.Context, errChan chan error) {
 	logrus.Debugf("mount watcher: opening %s", procMountInfo)
 	fd, err := unix.Open(procMountInfo, unix.O_RDONLY|unix.O_CLOEXEC, 0)
 	if err != nil {
-		logrus.Errorf("mount watcher: failed to open %s: %v", procMountInfo, err)
+		errChan <- fmt.Errorf("failed to open %s: %w", procMountInfo, err)
 		return
 	}
 	logrus.Debugf("mount watcher: opened fd=%d", fd)
@@ -33,7 +35,7 @@ func (u *Udev) watchMounts(ctx context.Context) {
 	}()
 
 	if fd > math.MaxInt32 {
-		logrus.Errorf("mount watcher: fd %d out of int32 range", fd)
+		errChan <- fmt.Errorf("fd %d out of int32 range", fd)
 		return
 	}
 	fdInt32 := int32(fd) //nolint:gosec // fd is guaranteed non-negative by Open and bounded by math.MaxInt32 check above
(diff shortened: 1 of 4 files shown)


pkg/udev/uevent.go
```diff
@@ -41,7 +41,7 @@ func (u *Udev) Monitor(ctx context.Context) {
 	// because any error will break the monitor loop.
 	errChan := make(chan error)
 	go u.spawnMonitor(ctx, errChan)
-
+	go u.watchMounts(ctx)
```
pohanhuang: What if something goes wrong in this go routine?
It will lost the the tracking the mount status, since there is no implementation like `spawnMonitor`

pkg/udev/uevent.go
pohanhuang, line 71: The original implementation keeps the monitor running but introduces a potential goroutine leak.

The problem is subtle: when `monitor-1` fails, it sends an error to `spawnMonitor`, which immediately launches `monitor-2` via `go u.monitor()`. But `monitor-1` hasn't finished its deferred cleanup yet — both goroutines are alive at the same time.

Under continuous failures, this compounds quickly:

```
t=1ms:   monitor-1 fails → monitor-2 starts
t=2ms:   monitor-2 fails → monitor-3 starts
t=3ms:   monitor-3 fails → monitor-4 starts
...
t=100ms: monitor-1's cleanup finally completes
```

By the time the first goroutine finishes, nearly 100 others have already been spawned. Under sustained failures, this becomes a goroutine bomb.

The fix is straightforward: run `fn` **_sequentially_** instead of concurrently. Since `delayFn.Until` calls `fn` synchronously, the next invocation only starts after the previous one fully returns — deferred cleanup included. At most one instance is ever running.

Benifit:
1. Ensure every monitor function to register through a uniform way
2. No needs to worry if the function early return and the monitor service is never rerun.

```go
// registerMonitorService runs fn sequentially with a fixed 1s retry interval
// until ctx is cancelled. Only one instance of fn runs at a time,
// preventing goroutine accumulation on repeated failures.
func registerMonitorService(ctx context.Context, name string, fn func(context.Context) error) {
    // Add one second wait to ensure the system get it data correct.
    backoff := &wait.Backoff{
        Duration: time.Second,
        Factor:   1.0,
        Steps:    math.MaxInt32,
    }
    delayFn := backoff.DelayWithReset(&clock.RealClock{}, 2*time.Minute)
    delayFn.Until(ctx, true, true, func(ctx context.Context) (bool, error) {
        if err := fn(ctx); err != nil {
            logrus.Errorf("%s failed: %v", name, err)
        }
        return false, nil
    })
    logrus.Infof("%s stopped", name)
}

func (u *Udev) Monitor(ctx context.Context) {
    go registerMonitorService(ctx, "udev-monitor", u.monitor)
    go registerMonitorService(ctx, "mount-watcher", u.watchMounts)
}
```

pohanhuang (changes requested):
Overall LGTM, have some concern about go routine leak.

pkg/udev/uevent.go
pohanhuang: inspired by https://www.nakamasato.com/kubernetes-training/kubernetes-operator/client-go/reflector/#overview.

However the backoff can be discussed.

Yu-Jack: Good idea. Let me rewrite it.

pkg/udev/uevent.go
Yu-Jack: Thanks for reminding. I missed this part.

pkg/udev/uevent.go
pohanhuang: After some thoughts.
I think we can further simplify the go routine to be 
```go

// registerMonitorService runs fn sequentially with a fixed 1s retry interval
// until ctx is cancelled. Only one instance of fn runs at a time,
// preventing goroutine accumulation on repeated failures.
func registerMonitorService(ctx context.Context, name string, fn func(context.Context) error) {
    // Add one second wait to ensure the system get it data correct.
    backoff := &wait.Backoff{
        Duration: time.Second,
        Factor:   1.0,
        Steps:    math.MaxInt32,
    }
    wait.BackoffUntil(func() {
        if err := fn(ctx); err != nil {
            logrus.Errorf("%s failed: %v", name, err)
        }
    }, r.backoffManager, true, ctx.Done())
    logrus.Infof("%s stopped", name)
}

func (u *Udev) Monitor(ctx context.Context) {
    go registerMonitorService(ctx, "udev-monitor", u.monitor)
    go registerMonitorService(ctx, "mount-watcher", u.watchMounts)
}
```

pohanhuang: See here for more: https://github.com/kubernetes/kubernetes/blob/6a5cc912a3c2a2c241fe05abea563079bf574abd/staging/src/k8s.io/apimachinery/pkg/util/wait/backoff.go#L226-L277

Yu-Jack: Done. Thanks!

pohanhuang, line 79: Should return error here?

pohanhuang, line 83: Should return error here,  en wait for the connect to stabilize.

pohanhuang (changes requested):
Overall LGTM, left with some minor error returns issue.

pkg/udev/uevent.go
Yu-Jack: done

Yu-Jack: done

pohanhuang (dismissed):
LGTM, thank you!

pkg/udev/mount_watcher.go
```diff
@@ -0,0 +1,75 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			logrus.Debug("mount watcher: context cancelled, exiting")
+			return nil
+		default:
+		}
+
+		logrus.Debugf("mount watcher: polling fd=%d for POLLERR, timeout=5s", fd)
+		n, err := unix.Poll(fds, 5*1000)
```
Vicente-Cheng: Why do we need to wake up every 5 seconds instead of a true busy wait?

pkg/udev/uevent.go
```diff
@@ -37,37 +38,38 @@ func NewUdev(opt *option.Option, scanner *blockdevice.Scanner) *Udev {
...
-func (u *Udev) spawnMonitor(ctx context.Context, errChan chan error) {
-	go u.monitor(ctx, errChan)
-	for {
-		select {
-		case err := <-errChan:
-			logrus.Errorf("failed to monitor udev events, error: %s", err.Error())
-			go u.monitor(ctx, errChan)
-		case <-ctx.Done():
-			return
+// registerMonitorService runs fn sequentially with a fixed 1s retry interval
```
Vicente-Cheng: This new function changes the behavior of the original. With this new function, we will have a 1s gap. If a udev event comes during that gap (the new monitor would not wake up), we might miss it.

pohanhuang: Great catch @Vicente-Cheng 
How about the following? I think this one does not change the behavior
```go
func registerMonitorService(ctx context.Context, name string, fn func(context.Context) error) {
    wait.UntilWithContext(ctx, func(ctx context.Context) {
        if err := fn(ctx); err != nil {
            logrus.Errorf("%s failed: %v", name, err)
        }
    }, 0) // 0 = restart immediately
}
```

pkg/udev/mount_watcher.go
Yu-Jack: I thought it was truly timeout stuff .. let me use `-1`.

Yu-Jack: updated

pkg/udev/uevent.go
Yu-Jack: Actually, original behavior still has `xx` ms delay because of restarting the monitor. 

Also, if you look into the source of event watcher, some errors don't break the loop, only skip it. However, we restart the monitor regardless of the type of errors.

```go
err := c.msgRead(buf)
if err != nil {
	errs <- fmt.Errorf("Unable to read uevent, err: %w", err)
	break loop // stop iteration in case of error
}

uevent, err := ParseUEvent(*buf)
if err != nil {
	errs <- fmt.Errorf("Unable to parse uevent, err: %w", err)
	continue loop // Drop uevent if not known
}
```

If we'd like to improve this, we might need two workers to listen the event. (We also need to resolve duplicated event issue)

Do you think we need to improve this in this PR? @Vicente-Cheng

Yu-Jack: Please check this commit to see my idea https://github.com/harvester/node-disk-manager/pull/317/changes/82addd047051a9ef67400ba75246ec4f1beacdab  cc @Vicente-Cheng @pohanhuang

pohanhuang: Thanks @Yu-Jack 
I think the safest way for now is to duplicate existing logic
```go
func (u *Udev) spawnMonitor(ctx context.Context, errChan chan error) {
	go u.monitor(ctx, errChan)
	for {
		select {
		case err := <-errChan:
			logrus.Errorf("failed to monitor udev events, error: %s", err.Error())
			go u.monitor(ctx, errChan)
		case <-ctx.Done():
			return
		}
	}
}
```
At least ensure the mount monitor is not stop, and create another issue to refactor the go routine, what do you think?

pohanhuang: The refactoring idea I have in mind is as follows.
The current implementation reopens the fd or connection on every restart. Ideally, we should establish the connection once and keep it alive across handler restarts to avoid the event loss window during reconnection.
I've drafted an initial implementation here, but I think it carries some risk to land right now. The safest approach for this PR is to duplicate the existing logic to keep the change minimal and reviewable. I hope that makes sense.

WDYT?
https://github.com/pohanhuang/node-disk-manager/pull/28/changes/29783a4ea8a9627cd5cdcb5bf9cdf01d0d9a35ad
cc @Vicente-Cheng @WebberHuang1118

Yu-Jack: It's good idea, and actually my commit also has the similar concept - reuse connection even receive the error from the monitor. 

> The safest approach for this PR is to duplicate the existing logic to keep the change minimal and reviewable

Yeah, I agree with you. I'll make the change later. For refactoring, we should open another issue to track instead of mixing them together to run into potential issue.

pohanhuang (dismissed):
LGTM

pohanhuang (approved):

Vicente-Cheng (approved):

pkg/udev/uevent.go
Vicente-Cheng: sounds good, thanks @pohanhuang, @Yu-Jack

Vicente-Cheng:
@mergifyio backport v1.9

PR #345, opened 2026-07-13
fix: always sync monut point instead of skipping

Yu-Jack:
**Problem:**

The original solution of https://github.com/harvester/node-disk-manager/pull/317 is a bit risky because it could delete the blockdevices.

**Solution:**

We just patch the mounted information back instead of removing the block device. Then, frontend can filter or mark that this device has been mounted somewhere.

Users can decide whether they'd like to unmount and use it or just ignore it.

Also, BTW, fix the CI issue.

**Related Issue:**
https://github.com/harvester/harvester/issues/8990
frontend: https://github.com/harvester/harvester-ui-extension/pull/987

**Test plan:**
image

### DIFF
--- a/pkg/controller/blockdevice/scanner.go
+++ b/pkg/controller/blockdevice/scanner.go
@@ -173,6 +173,9 @@ func (s *Scanner) handleExistingDev(oldBd *diskv1.BlockDevice, newBd *diskv1.Blo
 		oldBdCp.Status.DeviceStatus.Capacity = newBd.Status.DeviceStatus.Capacity
 		oldBdCp.Status.DeviceStatus.Details = newBd.Status.DeviceStatus.Details
 		oldBdCp.Status.DeviceStatus.Partitioned = newBd.Status.DeviceStatus.Partitioned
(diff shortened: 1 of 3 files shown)


.github/workflows/basic-ci.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
   job-new-installation:
     needs: validation
     concurrency:
-      group: ndm-new-installation
+      group: ndm-integration-tests
```
Yu-Jack: Use the same group to avoid some conflicts during testing.

Although we can use delay, but concurrency group is more reliable.

.github/workflows/basic-ci.yaml
```diff
@@ -119,7 +119,7 @@ jobs:
   jobs-upgrade:
     needs: validation
     concurrency:
-      group: ndm-upgrade
+      group: ndm-integration-tests
```
Yu-Jack: Same as the above.

Yu-Jack:
@mergify backport v1.9

Vicente-Cheng (approved):

